PR #6, opened 2020-06-01
Multithread Pixel lifters in Ops2 using LoopBuilder

gselzer:
This PR adds multi-threading support for lifting pixelwise Ops using imglib2's [LoopBuilder](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/loops/LoopBuilder.java) utility. As of this writing, we are only able to adapt `Function<I, O>` to `Function<RAI<I>, RAI<O>>` due to the limitations of `LoopBuilder`

Points of Discussion:
* Can we use `LoopBuilder` in other lifters in a reasonable way? (@maarzt would love your input here) I am thinking specifically about `IterableInterval`. This would (probably) require some change within `LoopBuilder`. 
  * It would be really cool if we could find some way to generalize `LoopBuilder` so that we could use it in scijava-ops as well, however I am not sure how possible that would be.
* Current adapt functionality is limited to arities 1-5 (since we only have Consumer Arities up to six [in `LoopBuilder`](https://github.com/imglib/imglib2/blob/b33186acc75abb8292da9865b6d100fc0787644d/src/main/java/net/imglib2/loops/LoopBuilder.java#L109)), however our `Function`s and `Computer`s reach much higher arities. Part of me thinks YAGNI.
* I did not include `Inplace` lifters. My thinking at the time was YAGNI. Opinions?


maarzt:
Hi @gselzer, `LoopBuilder` takes a lot of measures to run the code in a fashion that fits the Java just-in-time-compiler. It would unfortunately require a crazy amount of code to support all the Functions and Computers in Ops2 together with different combinations of RandomAccessibleIntervals and IterableIntervals.

There are actually several places that limit the arity in LoopBuilder:  [BindActionToSamplers](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/loops/BindActionToSamplers.java),  [FastCursorLoops](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/loops/FastCursorLoops.java), [SyncedPositionables](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/loops/SyncedPositionables.java)

I guess the lifters will become somehow similar to LoopBuilder. You could keep things a lot simpler if performance isn't your main concern. If performance is critical, than you would need to use [ClassCopyProvider](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/loops/ClassCopyProvider.java) to keep your lifters JIT-compilable. This is due to a problem discribed here: https://github.com/tpietzsch/none. You would need to write something like FastCursorLoops, for all your Functions and Computers.

gselzer:
Thanks for the reply @maarzt!

> It would unfortunately require a crazy amount of code to support all the Functions and Computers in Ops2

In order to support our higher-arity functional types, we would really only need higher arity `Consumer`s (i.e. a `SevenConsumer`, `EightConsumer`, etc..). All of our functional types could work inside that. I really don't know if we even need that though, since I cannot personally think of any `Function<RAI, RAI, RAI, RAI, RAI, RAI, RAI>` (that's 7x `RAI`) in our codebase. Our `Function.Arity7`s hardly ever have more than 3 `RAI`s in the `apply` method. So I am content for now with the maximum arity being 6.

> with different combinations of RandomAccessibleIntervals and IterableIntervals.

How difficult would it be to create `LoopBuilder` methods that work solely on `II`s (i.e. not a combination of `RAI` and `II`? It seems like all of these `RAI`s [are sometimes converted to `II`s anyways](https://github.com/imglib/imglib2/blob/b33186acc75abb8292da9865b6d100fc0787644d/src/main/java/net/imglib2/loops/LoopBuilder.java#L197)?

ctrueden:
Rebased over latest master, fixed compile errors, and merged! Thanks @gselzer!

PR #7, opened 2020-06-09
Set available pom-scijava version, make scripts executable

frauzufall:


ctrueden:
I think this is probably a vacuous merge with the commits I pushed independently doing the same things, but... we want GitHub to record that you also did it! :+1:

PR #9, opened 2020-06-12
Separate OpEnvironment and OpService

gselzer:
This PR tidies up the `OpService` and extracts much of the behavior into `DefaultOpEnvironment` (with exposed API belonging in `OpEnvironment`. 

The PR also rewrites calls to the `OpService`/`OpEnvironment` to reflect the changes in these files.


ctrueden:
Thanks!

PR #11, opened 2020-06-17
Make unionPluginInfo work on Windows

hinerm:


hinerm:
@ctrueden @gselzer I don't have any particular love for Groovy, it just has a dedicated execution plugin that works.

gselzer:
Hi @hinerm!

I checked out your branch and could not get scijava-ops to build when building from the incubator, since the paths to the `META-INF` files did not include `scijava/scijava-ops`. I think a nice solution would just be to use the absolute path to the `META-INF` files (via `System.getProperty('user.dir')`). This will only get you to the incubator level, though (unless you try to build from `scijava/incubator/scijava/scijava-ops`??). Unfortunately I do not have more time today to play with this.

hinerm:
Thanks @gselzer ! Whoops yeah I was just building from the `scijava/scijava-ops` directory directly. I'll fix that.

gselzer:
It might be worth checking out GMavenPlus. [This post](https://stackoverflow.com/questions/30051201/gmaven-plugin-how-to-set-property-in-pom-xml-for-external-groovy-script) suggests GMaven is no longer maintained, and GMavenPlus seems to have better support for passing command line args (which might be nice if we need this script in multiple projects). 

I am naive on this, but do you know if having Groovy installed is necessary to run this script? Or does GMaven take care of that for you?

hinerm:
>I am naive on this, but do you know if having Groovy installed is necessary to run this script? Or does GMaven take care of that for you?

That is a great question. You do **not** need to have Groovy installed. The plugin handles everything.

>It might be worth checking out GMavenPlus.

Agreed, nice catch!

hinerm:
OK, groovy did NOT work out because for whatever reason, groovyshell explodes if you try to do anything with a string with backslashes pulled in from Maven.

Ruby seemed to be OK with it though, so [new PR](https://github.com/scijava/incubator/pull/12)

PR #12, opened 2020-06-18
Fix pluginInfo merging on Windows

hinerm:


gselzer:
Nice work @hinerm! This works great! Builds from both a module level and from the incubator level (which I am sure you tested)!

My only consideration is about where the script should live. We do not want it to live in the incubator forever, so there are really two options:
1. Put it in pom-scijava
2. Put it in each project that needs it

I really don't know if I am familiar enough with pom-scijava to know if it belongs there. @ctrueden do you have opinions on this?

ctrueden:
Thanks @hinerm. Let's keep it here in incubator for now. Over time, we'll converge on a longer-term solution that can hopefully become part of pom-scijava.

pom.xml
```diff
@@ -99,5 +102,47 @@
...
+								<language>ruby</language>
+								<script>
+									# Append the source plugin annotations to the test plugin annotations
+									# Handle windows paths
+									basedir = '${project.basedir}'.gsub /\\+/, '\\\\'
+									sourceFile = basedir + '/target/classes/META-INF/json/org.scijava.plugin.Plugin'
+									destFile = basedir + '/target/test-classes/META-INF/json/org.scijava.plugin.Plugin'
+									input = File.open(sourceFile)
+									data_to_copy = input.read()
+									output = File.open(destFile, 'a')
```
imagejan, line 128: We could use Ruby's block syntax for `File.open` here, to ensure the file gets closed even in case of errors.

Something along these lines (sorry if it contains mistakes, I'm not a Ruby pro...):

```
File.open(destFile, 'a') { |f| f.write(data_to_copy) ) }
```

(and similar for the input file)

PR #15, opened 2020-06-30
Implement Op Lookup Caching

### LINKED ISSUE #33: Cache op lookups
Suggested by @tibuch: when looking up an op using an `OpRef` equal to a previous request, we should return a cached result, rather than going through the whole matching process again.

This is distinct from #5, which is about the possibility to cache op execution results. For this issue here, we just want to cache the matched op.

gselzer:
This PR implements a basic caching mechanism for storing the `Object`s returned by `Op` lookups. The underlying data structure is a `Map<OpRef, Object>`, where the `Object` is the `Op` returned by the matcher when `findOpInstance` was called with the corresponding `OpRef`.

TODO:
- [x] Consider size impact of cache. This cache *could* get rather large, considering the infinite potential of `OpRef`s. If not we would have to decide on a caching strategy and cache size: I think LRU makes sense, but we would have to implement a signal that tells the matcher whether or not to cache (and at that point, we might as well just implement scijava/scijava-ops#43). Otherwise `OpDependency` `OpRef`s will take over the cache.
- [x] Affirm that the changes within `OpRef.equals()`, `OpRef.hashCode()` are correct. This probably depends on decisions made within scijava/scijava-ops#36
- [x] Unit tests. Not sure how this can be tested. Maybe we could assert that the objects returned from two separate calls are equal, and are not if we tell the matcher not to cache (assuming scijava/scijava-ops#43 precedes this work)?

I did spend a little time thinking about scijava/scijava-ops#5, but I do not think that it belongs with this work. The matcher knows nothing about the `Object`s themselves during matching, so this is not the place to insert that kind of behavior.

Closes scijava/scijava-ops#33

### DIFF
--- a/scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
+++ b/scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
@@ -103,7 +103,9 @@ public class DefaultOpEnvironment extends AbstractContextual implements OpEnviro
 	 * Map to collect all aliases for a specific op. All aliases will map to one
 	 * canonical name of the op which is defined as the first one.
 	 */
-	private Map<String, Set<OpInfo>> opCache;
+	private Map<String, Set<OpInfo>> opDirectory;
+
+	private Map<OpRef, Object> opCache;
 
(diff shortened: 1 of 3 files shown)


ctrueden:
@gselzer Thanks!

PR #16, opened 2020-07-06
Clean up MatchingUtils

gselzer:
This PR introduces changes that clean up `MatchingUtils` and add missing support

TODO: 
- [x] Add support for `WildcardType`s in `MatchingUtils.inferTypeVariables()`

### DIFF
--- a/imagej/imagej-ops2/src/main/java/net/imagej/ops2/filter/ApplyCenterAwareNeighborhoodBasedFilter.java
+++ b/imagej/imagej-ops2/src/main/java/net/imagej/ops2/filter/ApplyCenterAwareNeighborhoodBasedFilter.java
@@ -29,11 +29,10 @@
 
 package net.imagej.ops2.filter;
 
-import net.imglib2.Cursor;
-import net.imglib2.IterableInterval;
-import net.imglib2.RandomAccess;
+import net.imglib2.RandomAccessible;
 import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.algorithm.neighborhood.Shape;
+import net.imglib2.loops.LoopBuilder;
 import net.imglib2.outofbounds.OutOfBoundsBorderFactory;
 import net.imglib2.outofbounds.OutOfBoundsFactory;
 import net.imglib2.view.Views;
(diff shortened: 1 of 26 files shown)


gselzer:
I am currently working on adding `WildcardType` support to [`MatchingUtils.inferTypeVariables`](https://github.com/scijava/incubator/blob/42a7503e87130a8692734c7f5b6b7a267d3a0e57/scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/MatchingUtils.java#L442), and have come across a conundrum:

The following scheme is legal (i.e. compiles and runs without error):
```java
public void caller() {
  List<Double> list = new ArrayList<Double>();
  list.add(1.5);
  Integer u = 1;
  callee(list, u);
}

public <T extends Number> void callee(List<? extends T> foo, T bar) {
  ...
}
```
In this scheme, we have `T` assigned to `Number` (as far as I can tell).

Now consider (suppose `<T extends Number>`):
```java
  final Type t = new Nil<T>() {}.getType();
  final Type listWild = new Nil<List<? extends T>>() {}.getType();
  final Type number = new Nil<Number>() {}.getType();
  final Type listDouble = new Nil<List<Double>>() {}.getType();
		
  final Type[] types = {listWild, t};
  final Type[] inferFroms = {listDouble, number};
		
  final Map<TypeVariable<?>, Type> typeAssigns = new HashMap<>();
  MatchingUtils.inferTypeVariables(types, inferFroms, typeAssigns);
```

We would thus expect to see the pair `(T, Number.class)` in `typeAssigns`, no?

How then do we ensure that `T` is mapped to `Number` within our current scheme? Upon encountering a `WildcardType`, we should probably scan the bounds of that wildcard, but what should any `TypeVariable`s within the bounds of that `WildcardType` be mapped to?

In the case above, we have `? extends T` and `Double`. Should `T` be mapped to `Double`? `Object`? I tend to think `Double`. But then we will get an `Exception` when we hit `T` and `Number`, because then `T` must be `Number` but has already been mapped to something else.

Note that we would also expect (I think?) `T` to be mapped to `Number` if the `Type number` was an `Integer` instead of a `Number`, however that seems more complex.

gselzer:
@ctrueden I am liking where this is, and am comfortable merging it.

I think we can also merge #17

ctrueden:
@gselzer Great. But I see WIPs in the history, still. Do you want me to squash and merge? Or do you want to try to fix the history? Either way is fine.

gselzer:
@ctrueden seeing as the point of this PR is to clean up a file, I am content with a squash and merge. Thanks!

PR #22, opened 2020-08-21
inferTypeVariables: insert Anys in Map when applicable

### LINKED ISSUE #49: Improve Handling of IAEs in DefaultOpEnvironment.adaptOp
[This line](https://github.com/scijava/incubator/blob/ba75e66182b07b5a8e5c2906a3af46e5fcf0431f/scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java#L406) is prone to throwing `IllegalArgumentException`s in calls such as

`ops.op("namespace.name").inType(Foo.class, Bar.class).apply()`

This call generates an `OpRef` looking for a `BiFunction<Foo, Bar, Object>`. If there is no satisfying `BiFunction`, adapt will try to look for an adaptor that can generate one. If [`adaptOpOutputSatisfiesRefTypes()`](https://github.com/scijava/incubator/blob/ba75e66182b07b5a8e5c2906a3af46e5fcf0431f/scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java#L387) can find an `adapt` Op to generate the appropriate `BiFunction`, the output `O` will not be assigned to a `Type` and thus `substituteTypeVariables` cannot substitute the `O` for some other `Type`.

Solutions include:
1. Modifying `inferTypeVariables` to put some dummy type (i.e. an `Any`) into the `Map`
2. Handling the error throwing within `substituteTypeVariables` better.

Unfortunately, I was rushed in thinking out this error report, so more details to come.

gselzer:
Closes scijava/scijava-ops#49

TODO:
* [x] Finish testing
  * [x] It would be nice to write an `Any.equals()` for testing purposes, however we must decide the semantics. When should two `Any`s be considered equal? Are there cases when other `Object`s (that are not `Any`s) should be considered equal?
  * [x] For which other `Type`s can this scenario occur? `Object` is special in that [`Types.getExactSuperType`](https://github.com/scijava/incubator/blob/762f117775b69cda61e6257595126e51b0c25073/scijava/scijava-types/src/main/java/org/scijava/types/Types.java#L4156) will return `null` (**not** `Object`) for a call `Types.getExactSuperType(i, Object.class)` for any interface `i` (since `Object` is not a supertype of `i`).

* [x] Clean `inferTypeVariables(ParameterizedType type, ...)`

### DIFF
--- a/scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/MatchingUtils.java
+++ b/scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/MatchingUtils.java
@@ -627,6 +627,8 @@ public final class MatchingUtils {
 					" cannot be implicitly cast to " + mappedType +
 					", thus it is impossible to infer type variables for " + inferFrom);
 			}
+			// for all remaining unmapped type vars, map to Any
+			mapTypeVarsToAny(type, typeMappings);
 		}
 		// -- edge cases -> do our best -- //
 		else if (superInferFrom == null) {
--- a/scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/MatchingUtils.java
+++ b/scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/MatchingUtils.java
@@ -612,9 +612,11 @@ public final class MatchingUtils {
 				// of type.
 				ParameterizedType superTypeOfType = (ParameterizedType) Types
 					.getExactSuperType(type, Types.raw(inferFrom));
-				if (superTypeOfType == null) throw new TypeInferenceException(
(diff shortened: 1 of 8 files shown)


gselzer:
After a bit of testing, I realized these errors come from attempting to infer type variables from:
* `Object`
* `Class`es and `Interface`s that have no type parameters (this doesn't come up often in scijava-ops, but could happen when trying to infer a `Img<O>` from an `Interval`)
* Raw `Class`es

These situations are all solved (well, the tests will fail until we solve the `Any` equals semantics)

gselzer:
@ctrueden I just rebased this over `master`, think it is ready to merge

PR #23, opened 2020-09-18
OpMethod support

### LINKED ISSUE #23: Add Op factory methods
Op methods are intended as a third way between Op fields and regular Op classes. They try to provide the conciseness of `@OpField`s while still enabling the use of other Ops via `OpDependency`s.
A typical implementation of an Op method may look like this:
```java
@Plugin(type = OpCollection.class)
public class MyOpCollection {

	@OpMethod(names = "test.multiplyNumericStrings")
	@Parameter(key = "numericString1")
	@Parameter(key = "numericString2")
	@Parameter(key = "multipliedNumericStrings", type = ItemIO.OUTPUT)
	@OpDependency(name = "test.parseInteger")
	public BiFunction<String, String, Integer> createMultiplyNumericStringsOp(
		final Function<String, Integer> parseIntegerOp)
	{
		return (i1, i2) -> parseIntegerOp.apply(i1) * parseIntegerOp.apply(i2);
	}
}
```
Here `createMultiplyNumericStringsOp` is a factory method ("Op method") that creates and returns the actual "test.multiplyNumericStrings" Op, whenever a new instance of this Op is requested by the user. The `@Parameter` annotations above the factory method describe the parameters of the returned Op itself, not the factory method. The `@OpDependency` annotation corresponds to the parameter of the factory method. The Op matching system takes care of resolving this dependency and passes it as an argument which can then be captured by the Op lambda.

The current progress on this issue is captured on branch [op-methods](https://github.com/scijava/scijava-ops/tree/op-methods), the relevant commits being b4b0d3bbf26cbf19f1cf8b0440cd6ef6fef8d50f through a9e860e6968a8290a80b7d0b157da0fa6661f391. The last of these commits is a (passing) test that further demonstrates the intended behavior.

In the long run, this way of injecting Op dependencies could make "ordinary" Op classes redundant. E.g., the above Op method could be a replacement for the following Op class:
```java
@Plugin(type = Op.class, name = "test.multiplyNumericStrings")
@Parameter(key = "numericString1")
@Parameter(key = "numericString2")
@Parameter(key = "multipliedNumericStrings", type = ItemIO.OUTPUT)
public class MultiplyNumericStrings implements BiFunction<String, String, Integer> {

	@OpDependency(name = "test.parseInteger")
	private Function<String, Integer> parseIntegerOp;

	@Override
	public Integer apply(String i1, String i2) {
		return parseIntegerOp.apply(i1) * parseIntegerOp.apply(i2);
	}
}
```
The benefit of that is that the factory method could also be used by ciients that don't want to use scijava-op's matching mechanism (given the method is public), simply by calling the method and manually providing a suitable Op as argument. In contrast, using the Op class at the bottom would require the use of reflection as there's no conventional access to the Op dependency field.
Also, Op methods could be thought of as an additional layer on top of POJOs to make them searchable and matchable, while keeping the business logic of the Op unaware of scijava-op entirely:
```java
// POJO, no dependencies on scijava-op. Required Op is injected via constructor.
public class MultiplyNumericStrings implements BiFunction<String, String, Integer> {

	private final Function<String, Integer> parseIntegerOp;

	public MultiplyNumericStrings(Function<String, Integer> parseIntegerOp) {
		this.parseIntegerOp = parseIntegerOp;
	}

	@Override
	public Integer apply(String i1, String i2) {
		return parseIntegerOp.apply(i1) * parseIntegerOp.apply(i2);
	}
}
```
```java
// scijava-ops-specific layer on top
@Plugin(type = OpCollection.class)
public class MyOpCollection {

	@OpMethod(names = "test.multiplyNumericStrings")
	@Parameter(key = "numericString1")
	@Parameter(key = "numericString2")
	@Parameter(key = "multipliedNumericStrings", type = ItemIO.OUTPUT)
	@OpDependency(name = "test.parseInteger")
	public BiFunction<String, String, Integer> createMultiplyNumericStringsOp(
		final Function<String, Integer> parseIntegerOp)
	{
		return new MultiplyNumericStrings(parseIntegerOp);
	}
}
```
Splitting the declaration and implementation of an Op like this would even allow us to write the declaration in some other language/file format than Java, if this proves beneficial.

gselzer:
This PR introduces support for writing ops as methods. Ops written as methods are discovered using the `@OpMethod` annotation, which declares both the `name` of the Op as well as the `type` of the Op (this allows for extensibility). Method ops allow the declaration of `OpDependencies`, which are included within the parameters accepted by the method. 

Method ops might look as follows:

```java
@OpMethod(names = "stats.mean", type = java.util.function.Function.class)
public Double meanOp(
  Iterable<Number> data, 
  @OpDependency(name = "stats.sum") Function<Iterable<Number>, Double> sumOp, 
  @OpDependency(name = "stats.size") Function<Iterable<Number>, Double> sizeOp) 
  {
    Double sum = sumOp.apply(data);
    Double size =  sizeOp.apply(data);
    return sum / size;
  }
```

This Op calculates the mean (hence the name `stats.mean`), and is a `Function<Iterable<Number>, Double>` (as declared by `type` and the set of parameters in the method). This Op also has two `OpDependencies`: one `Function<Iterable<Number>, Double> sumOp` and one `Function<Iterable<Number>, Double> sizeOp`. There are (should be :grin:) **no restrictions on the position of the `OpDependencies` within the method signature**.

The op would be called as follows:

```java
Iterable<Number> data = ...
Double mean = ops.op("stats.mean").input(data).outType(Double.class).apply();
```

Both `OpDependencies` will automagically be injected by the `OpService` using [Javassist](https://www.javassist.org/). **It is important to note that this PR does introduce a dependency on Javassist, which does not yet have a module**. When a method op has no `OpDependencies`, it will be constructed instead using [`LambdaMetaFactory`](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/LambdaMetafactory.html); this method is faster than using Javassist since we do not have to incur the compilation process during runtime.

This PR also introduces a suite of tests, ensuring that method Ops can be written for all of our supported Op types (`Computer`, `Inplace`, `Function`) **both with and without `OpDependencies`**.

TODO:
* [x] Decide whether we should proceed without a declared module name for Javassist. It would be nice to have one, but is probably not necessary.

Closes scijava/scijava-ops#23.

Thanks to @MarcelWiedenmann for starting this work.

### DIFF
--- a/scijava/scijava-ops/pom.xml
+++ b/scijava/scijava-ops/pom.xml
@@ -127,19 +127,20 @@
 			<artifactId>scijava-types</artifactId>
 			<version>${project.version}</version>
 		</dependency>
-
 		<!-- Third-party dependencies -->
 		<dependency>
 			<groupId>com.google.guava</groupId>
 			<artifactId>guava</artifactId>
 		</dependency>
-		
 		<dependency>
 			<groupId>io.leangen.geantyref</groupId>
 			<artifactId>geantyref</artifactId>
 			<version>${geantyref.version}</version>
 		</dependency>
-
+		<dependency>
+			<groupId>org.javassist</groupId>
+			<artifactId>javassist</artifactId>
+		</dependency>
 		<!-- Test scope dependencies -->
 		<dependency>
 			<groupId>org.junit.jupiter</groupId>
--- a/scijava/scijava-ops/src/main/java/module-info.java
+++ b/scijava/scijava-ops/src/main/java/module-info.java
@@ -22,4 +22,5 @@ module org.scijava.ops {
 
 	requires org.scijava;
 	requires org.scijava.types;
+	requires javassist;
 }
--- a/scijava/scijava-ops/src/main/java/org/scijava/ops/MethodParameterOpDependencyMember.java
+++ b/scijava/scijava-ops/src/main/java/org/scijava/ops/MethodParameterOpDependencyMember.java
@@ -1,0 +1,51 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
+ * Wisconsin-Madison and University of Konstanz.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+
+package org.scijava.ops;
+
+import java.lang.reflect.Parameter;
+import java.lang.reflect.Type;
+
+/**
+ * @author Marcel Wiedenmann
+ */
+public class MethodParameterOpDependencyMember<T> extends
+	AnnotatedOpDependencyMember<T>
+{
+
+	public MethodParameterOpDependencyMember(Parameter methodParameter,
+		final Type parameterType, OpDependency annotation)
+	{
+		// NB: "Real" parameter name may or may not be available during runtime,
+		// that is, the keys of instances of this class will likely be of the form:
+		// arg0, arg1, etc.
+		super(methodParameter.getName(), parameterType, annotation);
+	}
+}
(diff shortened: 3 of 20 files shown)


hinerm:
@gselzer is this something that needs to be merged before @ctrueden comes back?

gselzer:
@hinerm no, it is not urgent. I only pinged you to:
* Ensure this PR is not forgotten about
* Get a (third) opinion on the syntax for method Ops. As someone who is not particularly in the loop about scijava-ops, do you think this syntax could be improved?

hinerm:
@gselzer for me, this seems unfortunately long:
` @OpDependency(name = "stats.sum") Function<Iterable<Number>, Double> sumOp` 

If there was a way to do something like

```
@OpMethod(names = "stats.mean", type = java.util.function.Function.class, dependencies = {"stats.sum", "stats.size"})
public Double meanOp(
```

I think I might find that more pleasant, although I assume there's not a nice way to inject the parameters then.

Side note: what are the advantages of `@OpMethod` versus just taking `OpEnvironment` as a parameter, or giving all Ops implicit access to their environment? e.g. something like:
```
@OpMethod(names = "stats.mean", type = java.util.function.Function.class)
public Double meanOp(Iterable<Number> data) 
  {
    Double sum = ops().op("stats.sum").input(data).outType(Double.class).apply();
    Double size =  ops().op("stats.size").input(data).outType(Double.class).apply();
    return sum / size;
  }
```

Another side note: am I correct that in `OpBuilder` the op name doesn't actually get used until `apply` is called? If so it might be nice if the API allowed reusing the input and/or output configuration for multiple ops. Something like:

```
OpConfig conf = ops.op().input(data).outType(double.class);
Double sum = conf.apply("stats.sum");
Double size = conf.apply("stats.size");
```

gselzer:
Thanks for the feedback @hinerm. I will try to address each of your points:

> If there was a way to do something like

> ```java
> @OpMethod(names = "stats.mean", type = java.util.function.Function.class, dependencies = {"stats.sum", "stats.size"})
> public Double meanOp(
> ```

I agree that it is long, but it is also specific. We obviously cannot get shorter than `Function<Iterable<Number>, Double> sumOp` since that is the type and name of the parameter. This is already pretty long. 

Furthermore, we need to be able to determine whether a given parameter is an Op. We cannot assume to know that if a parameter is of type X, Y, or Z then it must be an `OpDependency`; this would not be extensible. This is compounded by the fact that many ops (such as [the `project` Op](https://github.com/scijava/incubator/blob/5a99c86c7360bc8f50ce100eebcacfd2ba17596b/imagej/imagej-ops2/src/main/java/net/imagej/ops2/project/DefaultProjectParallel.java#L57) ) take functions as parameters that **are not OpDependencies**. This behavior should be available for method-based Ops as well.

If you agree with what I said above (and I would be thrilled if you found a way around this), we must also include the `@OpDependency` annotation to those parameters, which will also require recording the name.

> Side note: what are the advantages of @OpMethod versus just taking OpEnvironment as a parameter, or giving all Ops implicit access to their environment?

The main benefit is that it provides independence from scijava-ops. If we wrote something along the lines of what you wrote above, we would then require anyone wanting to use our `OpMethod` to also require scijava-ops. @ctrueden has long been aware of people who do not want to go through the matching system, and this will allow them to do it. Of course, this requires them to find (write) their own sum and size ops in our example, so it is not particularly convenient to call them without Ops, but at least it can be done.

> Another side note: am I correct that in OpBuilder the op name doesn't actually get used until apply is called? If so it might be nice if the API allowed reusing the input and/or output configuration for multiple ops.

You are correct in that the op name is not used until `apply`/`compute`/`function`/`computer`/... is called. I agree that the construction is nice, however I would not want to use it for anything other than `Function`s; `Computer`s would not make sense under this construction since you would be overwriting the output again, and `Inplace`s could get really confusing really fast because you would have to keep track of which inputs were modified at which point. I am not sure how we would implement this for only `Function`s, thus I am hesitant to implement something along these lines. It might be worth further discussion, though...

PR #26, opened 2021-02-04
Scijava/scijava ops/hints

### LINKED ISSUE #43: Allow user to pass hints when calling Ops
It would be convenient for the user to give `OpService` some hints about how they want matching to behave in order to optimize the returns in the way that the user most wishes. For example, the user might want to:
* Prevent Op adaptation entirely
* Only allow Op adaptations that are not expensive, either in time or in space.
Note: this idea was mainly discussed in an Op adaptation context, however the possibilities with such a feature are endless and easily extensible

Ops could then declare certain hints that would then be examined by the matcher, possibly through a new `@OpHints` adaptation that Ops could implement.

gselzer:
This PR provides a two-way mechanism to deliver matching preferences (called hints) to the matcher. Hints can be declared by:
* Ops, using the [`OpHints`](https://github.com/scijava/incubator/blob/4efe3f8025a0bee0dea59d016f3e65376ed69120/scijava/scijava-ops/src/main/java/org/scijava/ops/hints/OpHints.java#L16) annotation. This data is stored within the `OpInfo`.
* Users, using the [`Hints`](https://github.com/scijava/incubator/blob/scijava/scijava-ops/hints/scijava/scijava-ops/src/main/java/org/scijava/ops/hints/Hints.java) interface. A `Hints` object can be created by the user or can be obtained through an `OpEnvironment`

These hints can then be used to indicate their matching preferences. For example, Ops can use the hint [`Simplification.FORBIDDEN`](https://github.com/scijava/incubator/blob/4efe3f8025a0bee0dea59d016f3e65376ed69120/scijava/scijava-ops/src/main/java/org/scijava/ops/hints/BaseOpHints.java#L16) to declare that it should not be simplified. Conversely, a user can pass the same hint to the matcher via a `Hints` object to specify that they do not want to match simplified Ops.

This PR also moves the simplification matching logic (`SimplifiedOpRef`/`SimplifiedOpInfo` construction, etc.) to `DefaultOpMatcher` (solves part of 

Closes scijava/scijava-ops#43.

TODO:
* [ ] Determine whether any other Hints should be supported out-of-the-box (@ctrueden @hinerm would you like to see any other hints available?)
* [x] Rebase over latest `master`


gselzer:
Assuming that we get `therapi-runtime-javadoc` working via scijava/scijava-ops#50, it might be worth considering a `Hint` to disable the javadoc scraping. Unsure if it is computationally expensive enough to warrant a hint, but definitely worth keeping an eye on.

scijava/scijava-ops/src/main/java/org/scijava/ops/OpInfo.java
```diff
@@ -29,9 +32,14 @@
 
 	/** Gets the associated {@link Struct} metadata. */
 	Struct struct();
-	
-	/** Describes whether this Op can be simplified. */
-	boolean isSimplifiable();
+
+	/** Gets the hints declared in the {@link OpHints} annotation */
+	List<String> declaredHints();
```
hinerm: Do we want to lose the connection to `OpHints` here? Is there any advantage to unwrapping it to a `List<String>`?

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpMethodInfo.java
```diff
@@ -78,7 +78,7 @@
 	private Struct struct;
 	private final ValidityException validityException;
 
-	private final boolean simplifiable;
+	private final List<String> hints;
```
hinerm: Same question - why a `List<String>` vs `OpHints`?

scijava/scijava-ops/src/main/java/org/scijava/ops/OpInfo.java
hinerm: OK, @gselzer: create a new `Hints` implementation that's read-only and backed by `ImmutableList`.

scijava/scijava-ops/src/main/java/org/scijava/ops/OpEnvironment.java
```diff
@@ -139,4 +147,16 @@ default OpBuilder op(final String opName) {
...
+	 * Sets the {@link Hints} for the {@link OpEnvironment}. Every Call to
+	 * {@link #op} that <b>does not</b> pass a {@link Hints} will <b>copy</b> the
+	 * Hints passed here (to prevent multiple Op calls from accessing/changing the
+	 * same {@link Hints}). In the case that <b>no {@link Hints} have been
+	 * given</b> to the Environment before {@code op} is called, the
+	 * implementation will create a suitable stand-in.
+	 * 
+	 * @param hints
+	 */
+	void setHints(Hints hints);
```
hinerm, line 221: TODO: add a disclaimer that to guarantee the `Hints` that are used during an op call, you should include the `Hints` in the call instead of at this environment level.

scijava/scijava-ops/src/test/java/org/scijava/ops/hints/AdaptationHintTest.java
```diff
@@ -0,0 +1,72 @@
...
+	public final Function<Double[], Double[]> op = (in) -> new Double[in.length];
+
+	@Test
+	public void testAdaptation() {
+		// make sure we can find the Op when adaptation is allowed
+		Hints hints = new DefaultHints();
+		hints.setHint(Adaptation.ALLOWED);
+		ops.env().setHints(hints);
+		@SuppressWarnings("unused")
+		Computers.Arity1<Double[], Double[]> adaptable = ops.op(
```
hinerm, line 30: @gselzer Could you add a `Hints` signature to the final methods of `OpBuilder` so that you can build an `Op` call with a particular set of `Hints`? (instead of requiring it to be set at the environment level)

scijava/scijava-ops/src/main/java/org/scijava/ops/OpInfo.java
gselzer: Done!

scijava/scijava-ops/src/main/java/org/scijava/ops/OpEnvironment.java
gselzer: Done!

scijava/scijava-ops/src/test/java/org/scijava/ops/hints/AdaptationHintTest.java
gselzer: Done!

scijava/scijava-ops/src/main/java/org/scijava/ops/OpInfo.java
hinerm: @gselzer I think we may want to return the actual `ImmutableHints` here, to concretely signal that the object is unmodifiable.  It's OK to have specific return types.

gselzer: @hinerm this is true, my only thought was that maybe there are times where it is too restrictive to return an `ImmutableHints`. But I can't think of a good use case, so I agree that we can make it `Immutable`

gselzer: @hinerm, ah, I forgot that the other reason for returning `Hints`: to return `ImmutableHints`, we'd have to `export` it. Is this something we want to do? Furthermore, since `ImmutableHints` extends `AbstractHints`, we'd have to export that as well, which I like less. I'd prefer to only expose the interfaces, what do you think of this?

hinerm: @gselzer that's fair

PR #27, opened 2021-02-20
Introduce separate Container concept for computer outputs

ctrueden:
Computer outputs aren't exactly mutable in the same way as inplace arguments are. Rather, they are preallocated empty containers whose contents will be filled in by the invoked computation.

My strong intuition is that having this distinction baked into the type annotations, and accessible at runtime, may be somehow helpful later.
    
This also adds a corresponding `ItemIO.CONTAINER` value for struct members; the `ItemIO.BOTH` value is retired in favor of `ItemIO.MUTABLE`, which corresponds to the `@Mutable` annotation for the `Inplace` interfaces.

### DIFF
--- a/scijava/scijava-ops/src/main/java/org/scijava/ops/function/Computers.java
+++ b/scijava/scijava-ops/src/main/java/org/scijava/ops/function/Computers.java
@@ -16,7 +16,7 @@ import java.util.function.Consumer;
 
 import org.scijava.ops.OpEnvironment;
 import org.scijava.types.Nil;
-import org.scijava.param.Mutable;
+import org.scijava.param.Container;
 import org.scijava.types.Types;
 
 /**
@@ -207,7 +207,7 @@ public final class Computers {
 		Consumer<O>
 	{
 
-		void compute(@Mutable O out);
+		void compute(@Container O out);
 
 		@Override
 		default void accept(final O out)
@@ -221,7 +221,7 @@ public final class Computers {
 		BiConsumer<I, O>
 	{
 
-		void compute(I in, @Mutable O out);
+		void compute(I in, @Container O out);
 
 		@Override
 		default void accept(final I in, final O out)
@@ -235,7 +235,7 @@ public final class Computers {
 		Consumers.Arity3<I1, I2, O>
 	{
 
-		void compute(I1 in1, I2 in2, @Mutable O out);
+		void compute(I1 in1, I2 in2, @Container O out);
 
 		@Override
 		default void accept(final I1 in1, final I2 in2, final O out)
@@ -249,7 +249,7 @@ public final class Computers {
 		Consumers.Arity4<I1, I2, I3, O>
 	{
 
-		void compute(I1 in1, I2 in2, I3 in3, @Mutable O out);
+		void compute(I1 in1, I2 in2, I3 in3, @Container O out);
 
 		@Override
 		default void accept(final I1 in1, final I2 in2, final I3 in3, final O out)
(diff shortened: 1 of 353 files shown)


scijava/scijava-ops/src/main/java/org/scijava/param/ParameterStructs.java
```diff
@@ -325,14 +324,16 @@ public static Method singularAbstractMethod(Class<?> functionalInterface) {
...
-	 * the return type of the method will me marked as {@link ItemIO#OUTPUT} and the
-	 * all method parameters as {@link ItemIO#OUTPUT}, except for parameters annotated
-	 * with {@link Mutable} which will be marked as {@link ItemIO#BOTH}. If the specified
-	 * type does not have a functional method in its hierarchy, null will be
-	 * returned.<br>
-	 * The order will be the following: method parameters from left to right, return type
+	 * Returns a list of {@link FunctionalMethodType}s describing the input and
+	 * output types of the functional method of the specified functional type. In
+	 * doing so, the return type of the method will me marked as
+	 * {@link ItemIO#OUTPUT} and the all method parameters as {@link ItemIO#OUTPUT},
```
imagejan, line 330:
```suggestion
	 * {@link ItemIO#OUTPUT} and the all method parameters as {@link ItemIO#INPUT},
```

scijava/scijava-ops/src/main/java/org/scijava/ops/function/Computers.java
```diff
@@ -431,7 +431,7 @@ default void accept(final I1 in1, final I2 in2, final I3 in3, final I4 in4, fina
 		Consumers.Arity17<I1, I2, I3, I4, I5, I6, I7, I8, I9, I10, I11, I12, I13, I14, I15, I16, O>
 	{
 
-		void compute(I1 in1, I2 in2, I3 in3, I4 in4, I5 in5, I6 in6, I7 in7, I8 in8, I9 in9, I10 in10, I11 in11, I12 in12, I13 in13, I14 in14, I15 in15, I16 in16, @Mutable O out);
+		void compute(I1 in1, I2 in2, I3 in3, I4 in4, I5 in5, I6 in6, I7 in7, I8 in8, I9 in9, I10 in10, I11 in11, I12 in12, I13 in13, I14 in14, I15 in15, I16 in16, @Container O out);
```
hinerm, line 434: Overall I think `@Mutable` and `@Container` are more descriptive, but I don't understand why some compute methods have an `@Container` but others don't?

e.g. https://github.com/scijava/incubator/pull/27/files#diff-199835d92f5a48202497fb4a82748776e19fe8d432e17c609dc462708ce36c11R558

gselzer:
@ctrueden as far as I can see, looks good to me!

I did see a minor issue with the `null` return on `ParameterStructs.findFunctionalMethodTypes`, so I added a commit to throw an `IAE` instead.

scijava/scijava-ops/src/main/java/org/scijava/ops/function/Computers.java
gselzer: @hinerm if I remember correctly, we only need these annotations on our functional interface methods (since this is where our API will pull them from). Your link goes to a method that is overriding the interface's method, thus it is probably best to leave the annotation off.

Do you have a rule for annotation application that you would prefer?

hinerm: @gselzer No that makes sense. I just noticed that the implementation methods had previously been annotated `@Mutable` so the drop in annotation stood out.

gselzer: Gotcha, thanks for pointing it out, but also that it is a good change (I think you agree with me on this, more just want to record my thoughts). I would personally want these annotations required (even allowable) in as few places as possible because it is:
* annoying to have to remember to put them there
* more work to change if/when we decide to alter the annotations/add new functional interfaces and support for them

ctrueden:
Shall we merge it?

ctrueden:
I merged it. :smile:

PR #28, opened 2021-03-25
Consolidate Parameter annotation data into Op Javadoc

gselzer:
This PR makes the usage of the `@Parameter` annotation obsolete; from now on, parameter names and descriptions should be described within the Op's Javadoc. At **compile time**, the Javadoc will be stored, to be parsed during runtime. As there are 3 different types of Ops, there are 3 different ways to store the parameter data:

# Ops written as Methods
Intuitively, the `@param` and `@return` tags can be applied within the Op's javadoc. **Note that if only some of the parameters are described, the parser will not be able to determine which parameters correspond to the given tags. In this case the parser will give up and synthesize names (and empty descriptions) for the Op**. This means that an Op written as
```java
	/**
	 * @param foo the first input
	 * @param bar the second input
	 * @return foo + bar
	 */
	@OpMethod(names = "test.javadoc.method", type = BiFunction.class)
	public static List<Long> OpMethodFoo(List<String> foo, List<String> bar) {
		BiFunction<String, String, Long> func = (s1, s2) -> Long.parseLong(s1) +
			Long.parseLong(s2);
		return Streams.zip(foo.stream(), bar.stream(), func).collect(Collectors
			.toList());
	}
```
will correspond to an `OpInfo`
```
public static java.util.List<java.lang.Long> org.scijava.param.JavadocParameterTest.OpMethodFoo(java.util.List<java.lang.String>,java.util.List<java.lang.String>)(
	 Inputs:
		java.util.List<java.lang.String> foo -> the first input
		java.util.List<java.lang.String> bar -> the second input
	 Outputs:
		java.util.List<java.lang.Long> output -> foo + bar
)

```

# Ops written as Classes
Documentation for Ops written as Classes is written similarly to that of `OpMethod`s, however the javadoc should be written on the *functional method* of the Op. For example:
```java
/**
 * Test Op used to see if we can't scrape the javadoc.
 * 
 * @author Gabriel Selzer
 */
@Plugin(type = Op.class, name = "test.javadoc.class")
class JavadocOp implements Function<Double, Double> {

	/**
	 * @param t the input
	 * @return the output
	 */
	@Override
	public Double apply(Double t) {
		return t + 1;
	}

}
```
This Op will produce an `OpInfo`:
```
org.scijava.param.JavadocOp(
	 Inputs:
		java.lang.Double t -> the input
	 Outputs:
		java.lang.Double output -> the output
)
```

# Ops written as Fields:
Ops written as `Field`s require different Javadoc syntax, as `Field`s do not have the `@param` and `@return` tags that `Method`s do (and we don't want to misuse them 😅 ). Thus we substitute `@param` for `@input` and `@return` for `@output`:
```java
	/**
	 * @input in the input
	 * @output the output
	 */
	@OpField(names = "test.javadoc.field")
	public final Function<Double, Double> javadocFieldOp = (in) -> in + 1;
```
This Op produces an `OpInfo`:
```
org.scijava.param.JavadocParameterTest.javadocFieldOp(
	 Inputs:
		java.lang.Double in -> the input
	 Outputs:
		java.lang.Double output -> the output
)
```
Ideally, we would write `Taglet`s for both `@input` and `@output`, however I have not (yet) had success in doing so (as the syntax changed in Java 9). This work *may* be pursued in a later PR.

TODO:
* [x] Do we pursue the `Taglet` work now, or at a later date (so we can get this out of the door). **There is some work on an `InputTaglet` in this iteration of the codebase. We should move it to another branch if we decide to wait on it**.
* [x] Clean Git History / decide to squash + merge (looking at you d2c131d)
* [x] Merge #35, as this work is built on top of that work

Notes:
* This PR introduces a dependency on [`therapi-runtime-javadoc`](https://github.com/dnault/therapi-runtime-javadoc). The version information should probably be moved upstream.


gselzer:
To set up javadoc scraping within your own Java 11 Maven project:
1) Depend on [`therapi-runtime-javadoc`](https://github.com/dnault/therapi-runtime-javadoc) and `javax.annotation-api` within your project:
```
	    <dependency>
		    <groupId>javax.annotation</groupId>
		    <artifactId>javax.annotation-api</artifactId>
		    <version>1.3.1</version>
	    </dependency>
	    <dependency>
		    <groupId>com.github.therapi</groupId>
		    <artifactId>therapi-runtime-javadoc-scribe</artifactId>
		    <version>0.12.0</version>
	    </dependency>
		<dependency>
			<groupId>com.github.therapi</groupId>
			<artifactId>therapi-runtime-javadoc</artifactId>
			<version>0.12.0</version>
		</dependency>
```
**NOTE**: depending on scijava-ops will automatically register all three of these dependencies

2) Register `JavadocAnnotationProcessor`  within the maven compiler plugin:
```
<plugin>
	<groupId>org.apache.maven.plugins</groupId>
	<artifactId>maven-compiler-plugin</artifactId>
	<version>3.8.1</version>
	<configuration>
		<annotationProcessors>
			<annotationProcessor>com.github.therapi.runtimejavadoc.internal.JavadocAnnotationProcessor</annotationProcessor>
		</annotationProcessors>
		<debug>true</debug>
		<optimize>true</optimize>
		<compilerArguments>
			<AaddGeneratedAnnotation>true</AaddGeneratedAnnotation>
			<Adebug>true</Adebug>
		</compilerArguments>
		<annotationProcessorPaths>
			<path>
				<groupId>com.github.therapi</groupId>
				<artifactId>therapi-runtime-javadoc-scribe</artifactId>
				<version>0.12.0</version>
			</path>
		</annotationProcessorPaths>
	</configuration>
</plugin>
```
For users of JPMS, there is an additional step: You must `require therapi.runtime.javadoc`, report that the module `uses javax.annotation.processing.Processor` and `open` and `export` all packages within your module to `therapi.runtime.javadoc`. Thus a module `foo` with packages `bar` and `baz` might have the following `module-info.java`:

```java
module foo {
  exports foo.bar to therapi.runtime.javadoc;
  exports foo.baz to therapi.runtime.javadoc;

  opens foo.bar to therapi.runtime.javadoc;
  opens foo.baz to therapi.runtime.javadoc;

  requires therapi.runtime.javadoc;

  uses javax.annotation.processing.Processor;
}
```

You **must** open and export to all projects you wish to scrape the javadoc of.

I have created a small example of how this works [here](https://github.com/gselzer/ops-therapi-test)

ctrueden (dismissed):
Thank you very much for working on this! :sparkles: I am excited to see the `@Parameter` annotation DIAF. :laughing: :fire: 

We discussed a bit some of the issues surrounding this work. Here are some thoughts, in no particular order:

1. Therapi is an source-time annotation processor. It might have access to the parameter names of methods, even though they are typically not stored into the bytecode. If so, we could map the `@param` names to the actual parameter names, in case the `@param` declarations are out of order, or some are missing, or some are spurious. This would be better than the current assumption that everything is there, and in same order as the method. However, it's not a blocker for merging this PR—it can always be fixed later. The question for then is: do we need to fork therapi, or write our own annotation processor with analogous logic, or what?
2. Relatedly, there is also the flag allowing to compile parameter names into the bytecode—we could add this to pom-scijava so that every project does this by default. It expands the bytecode a little, but in my tests, it was small (<3%? I don't remember exactly). But does Ops need it?
3. We should test whether `<scope>provided</scope>` works for the therapi-related dependencies. Certainly, we do not want otherwise-independent projects such as `imglib2-algorithm` to become polluted with therapi runtime dependencies; it is only ops that needs to depend on it at runtime, for reading the javadoc. Relatedly, we should test in both JPMS and non-JPMS scenarios. Curtis thinks provided scope is probably _not_ what we ultimately want, even if it works, because it's also a form of dependency pollution of sorts. What we really want is to add therapi to the _annotation processor_ classpath. This can be done in the `maven-compiler-plugin` configuration. [Here](https://stackoverflow.com/q/14322904/1207769) is an SO post about it; in a nutshell, you can add the deps to the compiler plugin's dependencies, instead of the _project_ dependencies.
4. In the case of mixed `@param` and `@input`/`@output`, therapi indexes things in such a way that the order is lost. We should consider either:
    1. Submitting an issue and/or PR to therapi adding a general-purpose getter of all the javadoc annotations in original order; or
    2. Fork therapi and merge its logic into scijava-indexer, so that everyone can have this functionality without incurring additional annotation-time dependencies.

As far as what actually needs to happen for this PR to merge:

* [x] Convert therapi usage to annotation processor classpath (i.e. maven-compiler-plugin classpath). Ensure it works.
* [x] Clean up git history.
* [x] File issues for any remaining concerns, as detailed above.
* [x] Remove support for `@input`/`@output`/etc. from `OpClass` and `OpMethod`. People need to always use `@param`/`@return` for those types of ops. Otherwise, the javadoc won't be fully correct. This also avoids dealing with issue (4) above.
* [x] Log a warning for any op that does not have properly javadoc'd parameters.
* [x] Fix the JPMS configuration: every package that we want javadoc for needs to be opened and exported to therapi. This raises a larger question of: is there automation we could introduce into pom-scijava easing the burden of `module-info.java` for our community? This is outside the scope of ops specifically, but would be super helpful for people.

gselzer:
> [x] Fix the JPMS configuration: every package that we want javadoc for needs to be opened and exported to therapi. This raises a larger question of: is there automation we could introduce into pom-scijava easing the burden of `module-info.java` for our community? This is outside the scope of ops specifically, but would be super helpful for people

I solved this issue by using Apache Velocity to autogenerate the `module-info`. Ideally, as @ctrueden mentions above, it would be ideal for us to do this in the `POM`, however it is a bit of a tricky problem. The [conventional method of adding opens via the `POM`](https://stackoverflow.com/questions/64920521/how-to-pass-add-opens-jdk-module-configuration-to-maven-test), however, does not offer a simple solution.



> * Remove support for `@input`/`@output`/etc. from `OpClass` and `OpMethod`. People need to always use `@param`/`@return` for those types of ops. Otherwise, the javadoc won't be fully correct. This also avoids dealing with issue (4) above.

This is also fixed. Interestingly, I found out that therapi treats these as equivalent:

```java
	/**
	 * Tests javadoc scraping of mutable taglet
	 * 
	 * @param foo the i/o argument
	 */
```

```java
	/**
	 * Tests javadoc scraping of mutable taglet
	 * 
	 * @param foo the i/o argument
	 * @return
	 */
```

As far as I can tell, there is no way to determine whether the `@return` tag has been written (although Eclipse *does* throw a warning in the latter case). In both scenarios, a return `Comment` is generated and is empty, so the best available method for checking whether or not there is a return tag is to check `doc.getReturn().toString.isEmpty()`. This isn't *quite* right, but is probably fine for right now. We might want to fix this if we make our own annotation processor / fork therapi

gselzer:
In response to the logging bullet point, I have created issue https://github.com/scijava/scijava/issues/64 (as it should sit on the backburner until https://github.com/scijava/scijava/issues/13 has been completed). For that reason, I will check that bullet point.

gselzer:
@ctrueden as of this writing, all commits build with passing tests

One consideration for a future issue: we still use Velocity to generate the `module-info` (and use `ModuleTest` to verify that packages are `open`ed to therapi correctly). But it would be better if we could have Maven do it for us automatically. But this is not feasible via `compilerArgs` alone as we'd have to hardcode in each package we want to open to a particular module. We'd want to do some research to see if a Maven plugin is capable of doing this for us. I **don't** think this is worth holding up the PR, though.

hinerm (approved):

imagej/imagej-ops2/pom.xml
```diff
@@ -210,7 +210,8 @@
 		<license.projectName>ImageJ software for multidimensional image processing and analysis.</license.projectName>
 
 		<!-- NB While we depend on old scijava-common -->
-		<allowedDuplicateClasses>org.scijava.*</allowedDuplicateClasses>
+		<imagej-ops2.allowedDuplicateClasses>${scijava.allowedDuplicateClasses}, org.scijava.*,com.github.therapi.runtimejavadoc.repack.com.eclipsesource.json.*</imagej-ops2.allowedDuplicateClasses>
```
ctrueden, line 213: Does the space before comma here cause any problems? If not, perhaps we should put a space before each comma, for readability? What about other whitespace? Newlines? If they also aren't an issue, we could avoid a super long line here in favor of a nice list of one entry per line.

scijava/scijava-ops/pom.xml
```diff
@@ -93,14 +103,56 @@
 
 		<license.licenseName>bsd_2</license.licenseName>
 		<license.copyrightOwners>SciJava developers.</license.copyrightOwners>
+		<enforcer.skip>true</enforcer.skip>
```
ctrueden, line 106: We cannot skip the enforcer checks. We need to find another way.

scijava/scijava-ops/pom.xml
```diff
@@ -93,14 +103,56 @@
...
 	<build>
 		<plugins>
 			<plugin>
 				<groupId>com.alexecollins.maven.plugin</groupId>
 				<artifactId>script-maven-plugin</artifactId>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<version>3.8.1</version>
```
ctrueden: The version is redundant here and should not be specified.

scijava/scijava-ops/pom.xml
```diff
@@ -93,14 +103,56 @@
...
 	</properties>
-
 	<build>
 		<plugins>
 			<plugin>
 				<groupId>com.alexecollins.maven.plugin</groupId>
 				<artifactId>script-maven-plugin</artifactId>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
```
ctrueden: The groupId is redundant here and should not be specified.

scijava/scijava-ops/pom.xml
```diff
@@ -93,14 +103,56 @@
...
+					<annotationProcessorPaths>
+						<path>
+							<groupId>com.github.therapi</groupId>
+							<artifactId>therapi-runtime-javadoc-scribe</artifactId>
+							<version>0.12.0</version>
+						</path>
+						<path>
+							<groupId>org.scijava</groupId>
+							<artifactId>scijava-common</artifactId>
+							<version>2.83.3</version>
```
ctrueden: Use `${scijava-common.version}` here, not `2.83.3`.

scijava/scijava-ops/pom.xml
```diff
@@ -93,14 +103,56 @@
...
+						</path>
+					</annotationProcessorPaths>
+				</configuration>
+			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
```
ctrueden: Redundant groupId.

scijava/scijava-ops/pom.xml
```diff
@@ -93,14 +103,56 @@
...
+				<artifactId>maven-jar-plugin</artifactId>
+				<configuration>
+					<archive>
+						<manifestEntries>
+							<Automatic-Module-Name>org.scijava.ops</Automatic-Module-Name>
```
ctrueden: Please use `<package-name>org.scijava.ops</package-name>` in the main `<properties>` section, rather than overriding the `maven-jar-plugin` configuration in this way. The pom-scijava-base configuration will propagate that value into the JAR manifest's `Automatic-Module-Name`.

scijava/scijava-ops/pom.xml
```diff
@@ -133,6 +185,22 @@
 			<groupId>org.javassist</groupId>
 			<artifactId>javassist</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>javax.annotation</groupId>
+			<artifactId>javax.annotation-api</artifactId>
+			<version>1.3.1</version>
```
ctrueden, line 151: Please declare `javax.annotation-api.version` as a property and use that here, rather than hardcoding.

scijava/scijava-ops/pom.xml
```diff
@@ -133,6 +185,22 @@
...
+		</dependency>
+		<dependency>
+			<groupId>com.github.therapi</groupId>
+			<artifactId>therapi-runtime-javadoc-scribe</artifactId>
+			<version>0.12.0</version>
```
ctrueden: Please declare `therapi-runtime-javadoc-scribe.version` as a property and use that here, rather than hardcoding.

scijava/scijava-ops/pom.xml
```diff
@@ -133,6 +185,22 @@
...
+		</dependency>
+		<dependency>
+			<groupId>com.github.therapi</groupId>
+			<artifactId>therapi-runtime-javadoc</artifactId>
+			<version>0.12.0</version>
```
ctrueden, line 156: Please declare `therapi-runtime-javadoc.version` as a property and use that here, rather than hardcoding. Since it appears that therapi is monoversioned, you can chain the properties like this:
```xml
<properties>
  ...
  <therapi.version>0.12.0</therapi.version>
  <therapi-runtime-javadoc.version>${therapi.version}</therapi-runtime-javadoc.version>
  <therapi-runtime-javadoc-scribe.version>${therapi.version}</therapi-runtime-javadoc-scribe.version>
  ... etc. ...
</properties>
```

scijava/scijava-ops/src/main/java/org/scijava/ops/create/CreateOpCollection.java
```diff
@@ -11,6 +11,11 @@
 import org.scijava.plugin.Plugin;
 import org.scijava.struct.ItemIO;
 
+/**
+ * Creation ops
+ * @author G
```
ctrueden, line 15: Hi G :wave: :smile:

scijava/scijava-ops/src/test/java/org/scijava/param/JavadocParameterTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.scijava.ops.core.OpCollection;
+import org.scijava.ops.create.CreateOpCollection;
+import org.scijava.ops.simplify.SimplificationUtils;
+import org.scijava.plugin.Plugin;
+import org.scijava.struct.ItemIO;
+import org.scijava.types.Types;
+
+/**
+ * Tests the ability of a Javadoc parser to scrape an Op's parameters out of its Javadoc
+ * @author G
```
ctrueden, line 32: Might want to grep for all of these.

scijava/scijava-ops/src/main/java/org/scijava/param/ParameterStructs.java
```diff
@@ -131,10 +131,21 @@ public static Struct structOf(final Class<?> c, final Method m)
 		// NB: Reject abstract classes.
 		checkModifiers(type.getName() + ": ", problems, type.getModifiers(), true, Modifier.ABSTRACT);
 
-		// Parse class level (i.e., generic) @Parameter annotations.
+		// obatin a parameterData (preferably one that scrapes the javadoc)
```
ctrueden: obtain

scijava/scijava-ops/src/main/java/org/scijava/ops/AnnotatedOpDependencyMember.java
```diff
@@ -74,6 +74,12 @@ public String getKey() {
 		return key;
 	}
 
+	@Override
+	public String getDescription() {
+		// TODO: consider more implementation here
```
ctrueden, line 79: What sort of implementation? This would be for describing an op dependency, right? Probably we should allow that, indeed. Someone might want to document the purpose of each particular op dependency. Or do I misunderstand?

scijava/scijava-ops/src/main/java/org/scijava/struct/Member.java
```diff
@@ -16,6 +16,9 @@
 	/** Unique name of the member. */
 	String getKey();
 
+	/** Description of the member. */
+	String getDescription();
```
ctrueden, line 20: I think I'd prefer `default String getDescription() { return ""; }` instead of abstract, here. The description is a thing that some struct implementations may not want to deal with. The original idea of the struct interfaces was to keep them kind of thin... while I do agree we need `getDescription()` to be a method here, but the default impl would give us the best of both worlds.

imagej/imagej-ops2/src/main/java/net/imagej/ops2/coloc/icq/LiICQ.java
```diff
@@ -53,11 +53,6 @@
  *            Type of the second image
  */
 @Plugin(type = Op.class, name = "coloc.icq")
-@Parameter(key = "image1")
```
ctrueden, line 56: The commit message looks suspiciously like something I would write. Did I do this work, or did you? Or both together? If I was involved, I should've been included in the commit message with a `Co-authored-by` line at the end of the description. Not worth it to rewrite history just for that, but for future reference.

scijava/scijava-ops/src/main/java/org/scijava/ops/OpUtils.java
```diff
@@ -382,7 +382,7 @@ private static String paramString(final Iterable<Member<?>> items, final Member<
 
 	/**
 	 * As {@link #paramString(Iterable, Member, String)} with a toggle to
-	 * control if inputs are types only or include the names.
+	 * control if inputs are types only or include the names/descriptions.
```
ctrueden: A small detail, but I think `names+descriptions` or similar would read more nicely than a slash.

scijava/scijava-ops/src/test/java/org/scijava/param/JavadocParameterTest.java
```diff
@@ -110,6 +126,101 @@ public void testJavadocMethodIR() {
...
+
+		// assert input names
+		String[] inputNames = info.inputs().stream().map(m -> m.getKey()).toArray(
+			String[]::new);
+		Assert.assertArrayEquals(inputNames, new String[] { "foo" });
+
+		// assert input descriptions
+		String[] inputDescriptions = info.inputs().stream().map(m -> m
+			.getDescription()).toArray(String[]::new);
+		Assert.assertArrayEquals(inputDescriptions, new String[] {
```
ctrueden, line 183: Actual and expected are reversed here, no?
I find it helps me get it right more often by declaring the expected explicitly on its own line, like this:
```java
var expected = {"the i/o argument"};
var actual = info.inputs().stream().map(m -> m.getDescription()).toArray(String[]::new);
Assert.assertArrayEquals(expected, actual);
```

scijava/scijava-ops/src/test/java/org/scijava/param/JavadocParameterTest.java
```diff
@@ -110,6 +126,101 @@ public void testJavadocMethodIR() {
...
+
+		OpInfo info = infos.next();
+		if (infos.hasNext()) {
+			Assert.fail("Multiple OpInfos with name \"test.javadoc.methodDependency\"");
+		}
+
+		// assert input names
+		String[] inputNames = info.inputs().stream().map(m -> m.getKey()).toArray(
+			String[]::new);
+		Assert.assertArrayEquals(inputNames, new String[] { "foo" });
```
ctrueden, line 207: Expected and actual reversed... probably we should go through all the tests (😬) to double check this. Maybe we can write some heuristics to semi-automate it.

scijava/scijava-ops/src/main/java/org/scijava/param/JavadocParameterData.java
```diff
@@ -175,6 +187,19 @@ private boolean hasVanillaJavadoc(MethodJavadoc doc, long numParams, long numRet
 		return sufficientParams && sufficientReturn;
 	}
 
+	/**
+	 * Determines whether {@code doc} has enough
```
ctrueden, line 188: It still worries me that we can only compare the _number_ here, because we don't have access to the parameter names. Would be much better to do keyword matching. Then if the number differs, you have what you have, and the other descriptions are just missing.

Do we have the parameter names at annotation processing time? If so, we could record them along with the javadoc. Would require changes to therapi, sure, but worth discussing after this work is merged. File an issue for it?

scijava/scijava-ops/src/main/java/org/scijava/param/JavadocParameterData.java
```diff
@@ -199,12 +224,29 @@ private boolean hasCustomJavadoc(List<OtherJavadoc> doc, long numIns, long numOu
...
+	 */
 	private void populateViaParamAndReturn(List<ParamJavadoc> params, Comment returnDoc) {
 		paramNames = params.stream().map(param -> param.getName()).collect(Collectors.toList());
 		paramDescriptions = params.stream().map(param -> param.getComment().toString()).collect(Collectors.toList());
 		returnDescription = returnDoc.toString();
 	}
 
+	/**
+	 * Populates {@link JavadocParameterData#paramNames},
+	 * {@link JavadocParameterData#paramDescriptions}, and, and
```
ctrueden, line 240: and, and

scijava/scijava-ops/src/main/java/module-info.java
```diff
@@ -1,29 +1,57 @@
...
+	opens org.scijava.ops.stats to therapi.runtime.javadoc;
+	opens org.scijava.ops.core to therapi.runtime.javadoc;
+	opens org.scijava.ops.provenance.impl to therapi.runtime.javadoc, org.scijava;
+	opens org.scijava.param to therapi.runtime.javadoc;
 
-  requires java.desktop;
+	// FIXME: This is a file name and is thus unstable
+	requires geantyref;
+
+	requires java.desktop;
```
ctrueden, line 49: I know this was already present, and thus tangential to this PR, but... why do we require `java.desktop`? We use none of those packages (`java.applet`, `java.awt`, `java.beans`, `javax.accessibility`, `javax.imageio`, `javax.print`, `javax.sound`, `javax.swing`), nor should we. If we have dependencies that do, that is a problem. File an issue for this for later?

scijava/scijava-ops/pom.xml
```diff
@@ -118,10 +118,6 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<annotationProcessors>
```
ctrueden: Ah, nice! If we don't need these anymore, do we even need the maven-compiler-plugin block in this POM at all anymore?

pom.xml
```diff
@@ -94,7 +94,14 @@
 			<plugin>
 				<artifactId>maven-compiler-plugin</artifactId>
 				<configuration>
+					<debug>true</debug>
+					<optimize>true</optimize>
 					<annotationProcessorPaths>
+						<path>
```
ctrueden, line 98: Ah, I guess we do need to keep the `annotationProcessPaths` section, and thus the `maven-compiler-plugin` block as a whole. But good that we can have it in the toplevel POM.

scijava/scijava-ops/pom.xml
```diff
@@ -113,17 +113,6 @@
 				<groupId>com.alexecollins.maven.plugin</groupId>
 				<artifactId>script-maven-plugin</artifactId>
 			</plugin>
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
```
ctrueden: Ah cool, ignore my previous comment about this, then!

ctrueden (commented):
All right, looks great! I made some comments, some of which accidentally became single comments rather than review comments—sorry for that. I also made some comments about things already addressed in later commits, since I was going commit by commit. Please bear with me. :smile:

gselzer:
> By `NSEException`, do you mean `NoSuchMethodError`? I was confused by that abbreviation. Not worth rewriting history over, though.

Likely I meant [`NoSuchElementException`](https://docs.oracle.com/javase/8/docs/api/java/util/NoSuchElementException.html).

imagej/imagej-ops2/pom.xml
gselzer: Fixed with 379e87cf

scijava/scijava-ops/pom.xml
gselzer: They were commented out at some point, and those comments are deleted as of ebd09714

scijava/scijava-ops/pom.xml
gselzer: Cannot determine which commit fixed this, but it is [definitely fixed](https://github.com/scijava/incubator/blob/ebd09714ef436abad2eec3d1398f0079f4b32ea7/pom.xml#L100) somewhere down the line

scijava/scijava-ops/pom.xml
gselzer: Cannot determine which commit fixed this, but it is [definitely fixed](https://github.com/scijava/incubator/blob/ebd09714ef436abad2eec3d1398f0079f4b32ea7/pom.xml#L100) somewhere down the line

scijava/scijava-ops/pom.xml
gselzer: Cannot determine which commit fixed this, but it is [definitely fixed](https://github.com/scijava/incubator/blob/ebd09714ef436abad2eec3d1398f0079f4b32ea7/pom.xml#L100) somewhere down the line

scijava/scijava-ops/pom.xml
gselzer: I'm unsure which commit commented out all of these, but I deleted the commented out lines [here](https://github.com/scijava/incubator/commit/ebd09714ef436abad2eec3d1398f0079f4b32ea7)

scijava/scijava-ops/pom.xml
gselzer: Again, not sure what commit made these fixes, but as of the [working end of this chain](https://github.com/scijava/incubator/blob/scijavva/scijava-ops-engine/view-javadoc-revisions/scijava/scijava-ops-engine/pom.xml) `package-name` is used instead.

scijava/scijava-ops/pom.xml
gselzer: Again, not sure what commit made these fixes, but as of the working end of this chain  the maven jar plugin is no longer used.

scijava/scijava-ops/pom.xml
gselzer: As of the [end of the working chain](https://github.com/scijava/incubator/blob/scijavva/scijava-ops-engine/view-javadoc-revisions/scijava/scijava-ops-engine/pom.xml) this dependency is no longer used.

scijava/scijava-ops/pom.xml
gselzer: fixed with 2f8267d3

scijava/scijava-ops/pom.xml
gselzer: Fixed with 2f8267d3

scijava/scijava-ops/src/main/java/org/scijava/ops/create/CreateOpCollection.java
gselzer: 👋

scijava/scijava-ops/pom.xml
gselzer: Whoops, I meant 976a700c

scijava/scijava-ops/pom.xml
gselzer: Whoops, 976a700c

scijava/scijava-ops/src/main/java/org/scijava/ops/create/CreateOpCollection.java
gselzer: Fixed with cfb18013

scijava/scijava-ops/src/test/java/org/scijava/param/JavadocParameterTest.java
gselzer: Think I got all of them in cfb18013

scijava/scijava-ops/src/main/java/org/scijava/param/ParameterStructs.java
gselzer: I guess I fixed when I created the `MemberParser` [implementations](https://github.com/scijava/incubator/blob/cfb18013919446ef230e8ef00081b4c12e4e356d/scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/struct/FieldParameterMemberParser.java#L27)

scijava/scijava-ops/src/main/java/org/scijava/ops/AnnotatedOpDependencyMember.java
gselzer: @ctrueden right, the question is: In what ways should the user be able to declare the description? Should they use some field within the `@OpDependency` annotation, or should we use javadoc? If we use javadoc, should we use a particular tag, support some custom taglet for it, or take the entire javadoc? I'm leaning towards taking the entire javadoc, as it would be the simplest option...

gselzer: Okay, so I made a fix in 67ef0552. I decided that for an `OpDependency` in a method, we can just use the `@param` tag to describe it. For an `OpDependency` as a `Field`, we can just take whatever is in the bulk javadoc.

Unfortunately, however, this seems to lengthen my incubator build by about 30 seconds. This inefficiency should be addressed, likely via lazily viewing the javadoc...

scijava/scijava-ops/src/main/java/org/scijava/struct/Member.java
gselzer: Fixed in  67ef055

imagej/imagej-ops2/src/main/java/net/imagej/ops2/coloc/icq/LiICQ.java
gselzer: Yes, if I remember correctly you wrote the initial script, which I then tweaked to produce the changes in this commit. I am sorry that your co-authorship did not persist 😦

scijava/scijava-ops/src/main/java/org/scijava/ops/OpUtils.java
gselzer: I am not able to find this method (or its javadoc) anymore \*shrugs\*

scijava/scijava-ops/src/test/java/org/scijava/param/JavadocParameterTest.java
gselzer: Fixed with deaa0497

scijava/scijava-ops/src/test/java/org/scijava/param/JavadocParameterTest.java
gselzer: Fixed with deaa0497

I don't know what kind of heuristics would be best here. Expected vs. actual seems more like a feel thing. I suppose we could look for `new` keywords or `String` literals? I would assume that this is a mostly-manual procedure though. Let's solve it in its own branch (this is noted in scijava/scijava#76)

scijava/scijava-ops/src/main/java/org/scijava/param/JavadocParameterData.java
gselzer: 😅 
Fixed with eafc09d2

scijava/scijava-ops/src/main/java/module-info.java
gselzer: It doesn't even seem worth its own issue...

Fixed with 07342c48

scijava/scijava-ops/pom.xml
gselzer: Short answer: [no](https://github.com/scijava/incubator/blob/07342c480541a31924d8f79ee247f8a9d531a6be/scijava/scijava-ops-engine/pom.xml#L113) 😁

pom.xml
gselzer: Yup!

scijava/scijava-ops/pom.xml
gselzer: Ignoring!

scijava/scijava-ops/src/main/java/org/scijava/param/JavadocParameterData.java
gselzer: > Do we have the parameter names at annotation processing time?

[This StackOverflow post](https://stackoverflow.com/questions/37319564/inspecting-parameter-names-from-annotation-processor) suggests so.

gselzer: I added this objective to scijava/scijava#65

scijava/scijava-ops/pom.xml
ctrueden: The commit that fixed it is f99bc5eb8e6855aca6f74a75cd2c26353acc5355. I determined this by using `git log -p` from the top of the branch, then searching for `maven-compiler-plugin` followed by pressing `n` to jump to each next match until I found the relevant one.

scijava/scijava-ops/pom.xml
ctrueden: You can use `git log -p` to list all the patches in a searchable form. Not that we really need to know which commit is responsible in this case, but just in general, it's a handy trick sometimes.

gselzer: :+1:

ctrueden:
I don't see anything else substantial to do on this PR, so I didn't even bother to start a review. My review points from earlier have already been addressed. Thanks! :100:

PR #29, opened 2021-04-16
Establish Scijava-functions as an independent module

gselzer:
This PR introduces a new module, `scijava-functions`, into the incubator. At the time of writing, this module contains only:
* The basic functional interfaces (`Computer`s, `Function`s, `Inplace`s, and `Producer`) used by Ops
* The annotations (`@Container`, `@Mutable`) designed to be used with those Ops
This partition is motivated by a desire to abstract the interfaces themselves from our particular use case. It is plausible that other developers might find our interfaces useful and would wish to use them without the baggage of Ops/the Ops matcher.

Points of discussion:
* [x] Is `scijava-functions` a suitable name? Can we think of a better one? It is a little confusing that the `Computer`s interfaces are found in `scijava-functions`...
* [x] Do the [`Functions.ArityXAsN`](https://github.com/scijava/incubator/blob/8c9fff5587efd25c83714252c4b979fe75c56b95/scijava/scijava-ops/src/main/java/org/scijava/ops/function/Functions.java#L1026) interfaces belong within `scijava-functions`? They have no dependence on/relationship to Ops, but would developers get any use out of them? We would also have to make them `public` (as they are currently `protected`)
* [x] The ops-related functionality within `Computers`/`Inplaces`/`Functions` has been repackaged within `ComputerUtils`/`InplaceUtils`/`FunctionUtils` (this is mostly referring to the `match()` methods. Should these classes continue to reside within `org.scijava.ops.function`, or should we move them into `org.scijava.ops.util`? We might also want to consider moving `OpWrappers`...
* [x] Do the annotations pertaining to Ops (`@OpMethod`/`@OpField`/`@OpDependency`, etc.) belong in this module as well? Anyone who wants to write Ops will need these annotations in addition to the interfaces enclosed in this module, however this module is spiritually independent of Ops; adding these annotations would remove that spirit.


ctrueden (changes requested):
This is awesome, and awesomely clean! :trophy:

Here's what we discussed should be changed:

* [x] For simplicity, and because this module is highly unlikely to grow in size, let's dispense with the `annotations` subpackage, and just keep all seven `.java` files at the top level.
* [x] The package name will become `org.scijava.function` (singular) for consistency with Java core's _functional interfaces_ package `java.util.function`.
* [x] Add a `package-info.java` explaining that the naming aligns with `java.util.function` and is all about functional interfaces, _not to be confused with_ SciJava Ops's use of the term "function". Relatedly, we discussed whether to use a different term in Ops now, to avoid a name clash here, but, ultimately decided that keeping the name "function" is the "least bad" approach for a number of reasons.
* [x] Add `ArityN` interfaces to container classes, but without the matching part—just the interface adaptation part. See below.

Points for future consideration:

* For `FunctionUtils` et. al matching methods: since they all are static but take an `OpEnvironment` as their first parameter, they are almost begging to be instance methods of `OpEnvironment` itself. The two main points against this are: 1) it greatly expands the `OpEnvironment`'s API (but I'd argue not so greatly that it overwhelms callers, but rather empowers them to more easily match functions etc.); and 2) it "bakes in" three core functional op types more deeply. We need to be careful with (2) to avoid resulting in a system that does not support extensibility into additional functional types. But it might be worth the tradeoff here...
* About `OpWrapper`: our thinking as of our discussion is that the `org.scijava.ops.function` package will go away, with the wrappers moved to join the matcher, because it's an implementation detail that we want to encapsulate, and is only used by the matcher.
* We decided not to move the op annotation interfaces (`OpClass`, `OpMethod`, `OpField`) into `scijava-function`, because they feel like bloat. However, they cannot live in `scijava-ops` as currently structured either, because then anyone who wants to use them will inherit bad dependencies (e.g. guava, javassist, therapy... 🤢) We agreed (reluctantly) that we need a separate Ops API module. The questions to ponder are: 1) how to name it? (`scijava-ops-api` and `scijava-ops-engine`?) 2) which classes exactly belong in API vs impl module? Reviewing all the logic currently in `scijava-ops` one by one will be needed.

<details><summary>Interface adaptation for n-ary</summary>

```java
Functions.Arity3<I1, I2, I3, O> myFunc = ....
Functions.ArityN<O> myFuncN = Functions.nary(myFunc);

public static Functions.ArityN<O> nary(Functions.Arity3<?, ?, ?, O> f) {
  return new ArityN<>() {
	Functions.Arity3<Object, Object, Object, O> func;

	public Arity3AsN(Functions.Arity3<Object, Object, Object, O> func) {
		this.func = func;
	}

	public O apply(Object... ins) {
		// Consider doing more safety checks here, but I'm initially inclined against...
		//if (ins[0] genericinstanceof func's first param type) and .... and length matches...
		return func.apply(ins[0], ins[1], ins[2]);
	}

	public Function getFunction() {
		return func;
	}
  }
}
public static Functions.ArityN<O> nary(Functions.Arity4<?, ?, ?, ?, O> f) { ... }
... code generated ...
```

</details>

gselzer:
Time commitment: 1 week

gselzer:
@ctrueden I have resolved all bullet points resulting from our discussion, and have also deleted the `org.scijava.ops.function` package. As of this writing, `ComputerUtils`/`FunctionUtils`/`InplaceUtils` have been moved to `org.scijava.ops.util`, and `OpWrappers` have been moved to `org.scijava.ops.matcher`. I think that our other two "points for further consideration" deserve their own issues (and the third point is likely covered by https://github.com/scijava/scijava/issues/59), so I think we are all good to go here?

gselzer:
As of this comment, all commits build with passing tests

hinerm (approved):

PR #31, opened 2021-05-20
Initialize scijava-struct as a new module

gselzer:
This PR introduces `scijava-struct`, a new module extracting the `Struct`/`Member` data structures from the present Ops module. Partitioning these data structures into their own module allows us to unify logic between SciJava Ops and KNIME (see https://github.com/scijava/scijava/issues/58)

TODO:
* There are currently zero tests; we should be testing these data structures
* We need to decide the package structure (do we need an `impl` package that is **not** `export`ed? Can we manage this with https://github.com/scijava/scijava/issues/62 in mind?
* Merge #28; that PR makes changes to `Member` (namely, adding a `description` `Field`); therefore, this PR is built on top of it


scijava/scijava-struct/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+		-->
+		<contributor>
+			<name>None</name>
+		</contributor>
+	</contributors>
+
+	<mailingLists>
+		<mailingList>
+			<name>Image.sc Forum</name>
+			<archive>https://forum.image.sc/tags/scijava-struct</archive>
```
ctrueden, line 67: Should be `/tag/scijava`.

scijava/scijava-struct/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+		</mailingList>
+	</mailingLists>
+
+	<scm>
+		<connection>scm:git:git://github.com/scijava/incubator</connection>
```
ctrueden, line 72: Needs to be `scm:git:https://github.com//scijava/incubator` since GitHub stopped supporting `git://`.

scijava/scijava-struct/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+		<tag>HEAD</tag>
+		<url>https://github.com/scijava/incubator</url>
+	</scm>
+	<issueManagement>
+		<system>GitHub Issues</system>
+		<url>https://github.com/scijava/scijava-struct/issues</url>
+	</issueManagement>
+	<ciManagement>
+		<system>Travis CI</system>
+		<url>https://travis-ci.com/scijava/incubator</url>
```
ctrueden, line 83: Needs to change to GitHub Actions.

scijava/scijava-struct/src/main/java/org/scijava/struct/DefaultStructInstance.java
```diff
@@ -0,0 +1,43 @@
+
+package org.scijava.struct;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class DefaultStructInstance<C> implements StructInstance<C> {
```
ctrueden: We should definitely javadoc the struct stuff. Of course we need javadoc on all public API, but struct is super abstract and will be very helpful to have it documented for later.

scijava/scijava-struct/src/main/java/org/scijava/struct/ItemIO.java
```diff
@@ -0,0 +1,80 @@
...
+	 * @see org.scijava.ops.function.Computers
+	 */
+	CONTAINER,
+	/**
+	 * The item is a hybrid input/output for the operation, whose current value will
+	 * be used as input to the computation, but subsequently overwritten with a
+	 * result.
+	 * 
+	 * @see org.scijava.param.Mutable
+	 * @see org.scijava.ops.function.Inplaces
```
ctrueden: struct does not depend on any scijava ops components, right? So this `@see` is not valid. It's great to point out that `MUTABLE` is relevant to inplace functions in ops, but I think we need to do it in a "non-type-safe" way, unfortunately.

scijava/scijava-struct/src/main/java/org/scijava/struct/ItemIO.java
```diff
@@ -0,0 +1,80 @@
...
+	/**
+	 * The item is an output for the operation.
+	 */
+	OUTPUT,
+	/**
+	 * The item is a hybrid input/output for the operation, whose structure is fixed
+	 * but whose content will be populated during execution of the operation.
+	 * 
+	 * @see org.scijava.param.Container
+	 * @see org.scijava.ops.function.Computers
```
ctrueden: This `@see` may not be valid; see below.

scijava/scijava-struct/src/main/java/org/scijava/struct/Member.java
```diff
@@ -0,0 +1,66 @@
...
+
+	/** Description of the member. */
+	String getDescription();
+
+	/**
+	 * Gets the type of the member, including Java generic parameters.
+	 * 
+	 * @see Field#getGenericType()
+	 */
+	// TODO: Use Type<T> or Nil<T> from new scijava-types.
```
ctrueden: This TODO is obsolete now, right? Or should this be `Nil<T>`?

scijava/scijava-struct/src/main/java/org/scijava/struct/Member.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	default Struct childStruct() {
+		return null;
+	}
+
+	default MemberInstance<T> createInstance(
+		@SuppressWarnings("unused") Object o)
+	{
+		return () -> Member.this;
```
ctrueden: My brain is probably not working today, but... what even is this syntax‽

Edit: OK, I get it... it's (mostly) equivalent to:
```java
Member<T> me = this; return new MemberInstance<T>() { @Override Member<T> member() { return me; } }
```
but in slick lambda shorthand.

ctrueden (approved):
I wrote this code together with Christian over many painstaking iterations at a hackathon, so it's not surprising that I'd be satisfied with it as is. But I am. Satisfied. 😆

PR #33, opened 2021-06-07
Initialize SciJava Concurrent module

### LINKED ISSUE #31: Determine a multithreading solution
Most existing Ops rely on [`DefaultChunker`](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/thread/chunker/DefaultChunker.java) when they wish to multithread a process. This process is more complicated to introduce in scijava-ops because:
1) It requires all ops that want to multithread to require an `ExecutorService` as a parameter, something that could/should be injectable by the `OpService`.
2) The tests for this `Op` introduce `junitbenchmark`; it would be nice if we did not have to introduce this.

There are a few different approaches that we can take:
1) Create some sort of static method that mimics the behavior of `Chunk`
2) Add support for what is going on in https://github.com/imglib/imglib2-algorithm/issues/81

Would it be possible to create a special mutlithreading `OpDependency` that, upon discovery, the matcher could inject its `ThreadService` into? i.e. if you had a
```
@OpDependency(name = "thread.chunker")
private NullaryComputer<Chunk> chunkerOp;
```
Ops could find a `Computer` that takes an `ExecutorService` as input and wrap that into a `NullaryComputer` by injecting its `ExecutorService`?

### LINKED ISSUE #120: Desirable Ops conveniences
- [ ] `Shapes` specifiable with a single number
- [ ] Not needing to provide an [ExecutorService](https://github.com/scijava/incubator/blob/6dc6a92a479edffff1a0544768d41d78c4ef89cd/imagej/imagej-ops2/src/main/java/net/imagej/ops2/filter/gauss/Gaussians.java#L74-L80)

Once I got to the `ExecutorService` I realized that ImageJ Ops2 probably needs a pass updating which Ops its wrapping. And actually many of those will probably just be cut by directly wrapping [imglib2-algorithms](https://github.com/scijava/scijava/issues/56). For example, the `filter.gauss` method in the above case is now deprecated and superced by [this method](https://github.com/imglib/imglib2-algorithm/blob/d36c5204ad982bdbe687dea3c62790ddb7e71a98/src/main/java/net/imglib2/algorithm/gauss3/Gauss3.java#L136).

### LINKED ISSUE #125: Add Lifting to Neighborhood Types

gselzer:
This PR migrates @maarzt's fantastic parallelization work into a library independent of imglib2 - allowing us to use his work in a broader context throughout other SciJava libraries. This work is a culmination of the discussion in https://github.com/scijava/scijava/issues/31

TODO:
* [x] Is there a better name? As @ctrueden noted in https://github.com/scijava/scijava/issues/31, `scijava-parallel` is taken. We might want to do some renaming if we keep the name `scijava-concurrent`
* [x] @maarzt - can you add to the POM the `developer` roles you feel comfortable with? If you want add yourself as `author` in all of the classes, that'd be great too. I think the only one I changed significantly is `ParallelizationBenchmark.java`...
* [x] Similarly, @ctrueden you have full roles since I copied the POM from SciJava Types 😝. Do you want these?
* [x] We should decide if there are any classes in this module that should not be `export`ed. I think the package is small enough where it is probably unnecessary, but we might want to hide `ForkJoinExecutorService`/`SequentialExecutorService` away in some `org.scijava.concurrent.impl` package.

Closes https://github.com/scijava/scijava/issues/31


gselzer:
I am unable to run `ParallelizationBenchmark.main` within Eclipse, getting 

```
WARNING: Unknown module: org.scijava.concurrent specified to --patch-module
WARNING: Unknown module: org.scijava.concurrent specified to --add-reads
Error: Could not find or load main class org.scijava.concurrent.ParallelizationBenchmark
Caused by: java.lang.ClassNotFoundException: org.scijava.concurrent.ParallelizationBenchmark
```

Can anyone reproduce this?

imagej/imagej-ops2/src/main/java/net/imagej/ops2/coloc/pValue/DefaultPValue.java
```diff
@@ -89,46 +89,38 @@ public void compute(final RandomAccessibleInterval<T> image1, final RandomAccess
...
+				if (i > 0)
+					shuffled.shuffleBlocks(seeds[index]);
+				copy(shuffled, buffer);
+				sampleDistribution[index] = op.apply(buffer, trimmedImage2);
 			}
-		} catch (final InterruptedException | ExecutionException exc) {
+		};
+
+		try {
+			Parallelization.getTaskExecutor().forEach(params, task);
```
hinerm: @gselzer Didn't you get the `executor` above on line 92?

gselzer: Whoops, good catch! Fixed in f992fe5

gselzer:
> We should decide if there are any classes in this module that should not be exported. I think the package is small enough where it is probably unnecessary, but we might want to hide ForkJoinExecutorService/SequentialExecutorService away in some org.scijava.concurrent.impl package.

I think the answer to this is probably no - it seems to me like the user would want to choose their `ExecutorService` implementation, and they couldn't do that if we hid the implementations away. It's probably best to leave them exposed.

hinerm:
>We might want to do some renaming if we keep the name scijava-concurrent

@gselzer concurrent seems fine to me. The whole `scijava-parallel` thing is unfortunate. What would you rename though if we stick with concurrent?

gselzer:
> > We might want to do some renaming if we keep the name scijava-concurrent
> 
> @gselzer concurrent seems fine to me. The whole `scijava-parallel` thing is unfortunate. What would you rename though if we stick with concurrent?

I have no idea, it was too long ago - maybe I meant classes? Maybe we should rename `Parallelization` to `Concurrency`?

imagej/imagej-ops2/src/main/java/net/imagej/ops2/filter/convolve/ConvolveNaiveF.java
```diff
@@ -102,9 +102,9 @@ public RandomAccessibleInterval<O> apply(final RandomAccessibleInterval<I> input
 			final RandomAccessibleInterval<K> kernel, @Optional OutOfBoundsFactory<I, RandomAccessibleInterval<I>> obf,
 			final O outType) {
 
-		// conforms only if the kernel is sufficiently small
-		if (Intervals.numElements(kernel) <= 9)
-			throw new IllegalArgumentException("The kernel is too small to perform computation!");
+//		// conforms only if the kernel is sufficiently small
```
hinerm, line 105: @gselzer why is this commented out and not just deleted?

scijava/scijava-concurrent/pom.xml
```diff
@@ -53,45 +53,63 @@
...
 	</scm>
 	<issueManagement>
 		<system>GitHub Issues</system>
-		<url>https://github.com/scijava/scijava/issues</url>
+		<url>https://github.com/scijava/scijava-concurrent/issues</url>
 	</issueManagement>
 	<ciManagement>
-		<system>GitHub Actions</system>
-		<url>https://github.com/scijava/incubator/actions</url>
+		<system>Travis CI</system>
```
hinerm: @gselzer nope lol

scijava/scijava-threads/README.md
hinerm, line 1: @gselzer should probably have a basic readme in the new concurrent module?

imagej/imagej-ops2/src/main/java/net/imagej/ops2/filter/convolve/ConvolveNaiveF.java
hinerm: @gselzer why is this commented out and not just deleted?

hinerm (changes requested):
I love this change.. getting rid of the `ExecutorService` is fantastic and the consuming code looks simpler in many cases.

Just a couple minor points. Thanks @gselzer!

ctrueden:
Not a full review from me, but I just wanted to chime in that I'm also pleased to see the explicit `ExecutorService` parameters go away in favor of the new `Parallelization` paradigm that tucks it neatly away while still enabling configurability. Thanks for working on it!

hinerm:
@gselzer looking good. Let's not rename things since this is a port of ImgLib2 code, basically.

Per our talk this morning we'll wait to merge until you add a tutorial in `imagej2-ops2-tutorial` showing its use

hinerm:
> We should decide if there are any classes in this module that should not be exported.

I wouldn't call these internal API so I am fine exporting everything in this module

gselzer:
@hinerm are all of your requested changes complete now?

hinerm (approved):

PR #35, opened 2021-06-18
Data provenance mechanism

### LINKED ISSUE #53: Construct a mechanism for data provenance
As Ops grows, we aspire to wrap the algorithms of other libraries into callable ops. This can result in numerous implementations of many algorithms, and it is of interest to determine which implementation was used for a particular Op call.

There are a few ways that we could try to implement this (note that this is an incomplete list):
* @ctrueden suggested baking provenance data into "simplified" types (these types would arise as a part of the simplification work being done to solve #4)
* We might be able to bake this work in to the Op wrappers. This has the added benefit of recording data when no simplified types are used, but might extend past the intended purpose of the wrappers. 

see scijava/scijava-common#227

gselzer:
This PR introduces a new mechanism, `OpHistory`, to provide [data provenance](https://en.wikipedia.org/wiki/Data_lineage) for Ops and Op outputs involved with the matcher. This `OpHistory` provides a **static** repository of information about Op executions. For each **top-level** call to the matcher, a universally-unique identifier ([UUID](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html)) is created. Then, Ops stores:
* A `Graph<OpInfo>` describing the hierarchy of Ops involved in every return produced by the `DefaultOpEnvironment`
* The top-level Op returned by the call to the matcher, along with the `OpWrapper` result (if used)
* A history of the executions performed by the Op (when the `OpWrapper` was utilized), along with the output produced

Using these recorded Objects, we are able to then get answers for the questions:
* `OpHistory.executionsUpon(Object o)`: For a given `Object` output, what Op(s) were responsible for creating/mutating this `Object`?
* `OpHistory.opExecutionChain(Object op)`: For a given `Object` op, which `OpInfo`(s) were tasked with creating/populating the `OpDependency` needs to create this `Object`?

The answers to these questions allow us to obtain the `OpInfo`s (and thus the backing algorithms) needed to construct an Op output in a reproducible way.

This Op also refactors the outputs of various `DefaultOpEnvironment` to better facilitate the population of the `OpHistory`. When `DefaultOpEnvironment.op` is called, the general scheme that should be followed is:
1. match the Op
2. store the Op along with its corresponding `OpInfo` in the op cache
3. ask the cache for the Op, wrap the Op, and return the Op to the user

This new process allows us to store the "raw" Op in the cache, instead of the wrapped Op, better facilitating Op cache hits.

TODO:
* [x] Remove [this `HACK`](https://github.com/scijava/incubator/blob/9f32ec2a1fd5f184b04a7ce98c6a44e23e5b53c3/scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java#L151), determine when/how the history should be reset
* [x] Form an issue about persisting the `OpHistory`, possibly using https://github.com/scijava/incubator/pull/30. The usage of `UUID` should aid in the storage.
* [x] Determine what API pertaining to logging Op history (if any) belongs in `OpEnvironment`. Should data provenance be something that all `OpEnvironment`s implement, or something limited to our `DefaultOpEnvironment`?
* [x] Consider the return of a `Graph<OpInfo>` for `OpHistory.opExecutionChain`: this is exposing Google Guava data structures, which is not ideal. We need to be able to return a Tree of some sort, and this seems to be the best answer. 

Closes https://github.com/scijava/scijava/issues/53


hinerm:
> Remove this HACK, determine when/how the history should be reset

Possible solution: make the `OpHistory` a service so that each time we create a `Context` we get a fresh history

gselzer:
> Possible solution: make the OpHistory a service so that each time we create a Context we get a fresh history

@hinerm thanks for the suggestion, I just implemented it.

gselzer:
> Determine what API pertaining to logging Op history (if any) belongs in OpEnvironment. Should data provenance be something that all OpEnvironments implement, or something limited to our DefaultOpEnvironment?

If `OpHistory` is a Service, then we do not need to access that API through `OpEnvironment`, we can just directly access the `OpHistoryService` through the `Context`.

> Consider the return of a Graph<OpInfo> for OpHistory.opExecutionChain: this is exposing Google Guava data structures, which is not ideal. We need to be able to return a Tree of some sort, and this seems to be the best answer.

For lack of a better option, I say we keep this for now. @hinerm @ctrueden if you have other thoughts, feel free to express them.

Otherwise, let me know what you think about merging @hinerm

scijava/scijava-ops/src/main/java/org/scijava/ops/OpEnvironment.java
```diff
@@ -213,7 +213,7 @@ default OpBuilder op(final String opName) {
 	 * Note that this method does not inherently provide <b>any</b> guarantees
 	 * pertaining to thread-safety; this API is provided purely for convenience.
 	 * Any calls to {@link OpEnvironment#op(String, Nil, Nil[], Nil)} that require
-	 * a specific {@Hints} should <not> use this method, instead opting for
+	 * a specific {@Hints} should <b>not</b> use this method, instead opting for
```
ctrueden, line 216: Below, the method `setHints` should be `setDefaultHints`, since it's a fallback for when no hints are passed specifically. We also need a method `Hints getDefaultHints()` for the default ones, and `Hints createHints(String... startingHints)` or perhaps merely `Hints hints(String... startingHints)`.

scijava/scijava-ops/src/main/java/org/scijava/ops/conversionLoss/LossReporterWrapper.java
```diff
@@ -35,7 +45,15 @@
 			@Override
 			public Double apply(Nil<I> from, Nil<O> to) //
 			{
-				return op.apply(from, to);
+				// Call the op
+				Double output = op.apply(from, to);
+
+				// Log a new execution
+					if (!hints.containsHint(DependencyMatching.IN_PROGRESS)) {
```
ctrueden, line 52: Indentation

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/BaseOpHints.java
```diff
@@ -24,4 +24,12 @@
 		public static final String IN_PROGRESS = prefix + ".IN_PROGRESS";
 	}
 
+	// TODO: better naming
+	public static class DependencyMatching {
+		public static final String prefix = "dependencyMatching";
```
ctrueden, line 29: The prefix variable should be in capital case `PREFIX` because it's a constant.

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/Hints.java
```diff
@@ -20,6 +33,25 @@
...
+	/**
+	 * Generates a new {@link Hints} with identical hints.
+	 * 
+	 * @param generateID designates whether the returned {@link Hints} should
+	 *          designate a new execution chain ID, or whether it should maintain
+	 *          the ID of this {@link Hints}
+	 * @return a new {@link Hints} Object with the same hints as this
+	 *         {@link Hints}
+	 */
+	public Hints getCopy(boolean generateID);
```
ctrueden, line 45: This should not be a getter, because it creates things. Should be `copy(boolean)`.

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/Hints.java
```diff
@@ -20,6 +33,25 @@
...
+
+	/**
+	 * Returns the {@link UUID} uniquely identifying the an associated
```
ctrueden, line 48: Javadoc typo `the an`

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/AbstractHints.java
```diff
@@ -14,17 +15,34 @@
...
 	// Hints are stored by their hint type (the middle term)
 	final Map<String, String> hints;
 
-	public AbstractHints(String... startingHints) {
+	public AbstractHints(final String... startingHints) {
+		hints = new HashMap<>();
+		for(String hint : startingHints)
+			setHint(hint);
+
+		this.historyHash = UUID.randomUUID();
```
ctrueden, line 28: Instead of duplicating this code block, better to write:
```java
this(UUID.randomUUID(), startingHints);
```
Assuming you need to keep this constructor at all.

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/AbstractHints.java
```diff
@@ -14,17 +15,34 @@
...
+	}
+
+	public AbstractHints(final UUID historyHash, final String... startingHints) {
 		hints = new HashMap<>();
 		for(String hint : startingHints)
 			setHint(hint);
+		this.historyHash = historyHash;
```
ctrueden, line 35: What you could do is:
```java
this.historyHash = historyHash == null ? UUID.randomUUID() : historyHash;
```
Then you wouldn't have to repeat the UUID randomization code downstream.

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/AbstractHints.java
```diff
@@ -72,4 +90,9 @@ public int hashCode() {
 		return hints.hashCode();
 	}
 
+@Override
```
ctrueden, line 93: In `equals` above, either: let two hints be equal if they are both `Hints` (not `AbstractHints`), OR move this logic into `DefaultHints` and test that the other thing is also a `DefaultHints`, and document it in the `equals` javadoc. My weak preference would be to accept any `Hints` here, just for simplicity. A third option is to return false if `this.getClass() != that.getClass()`, forcing exactly the same implementation, which ensures subclasses don't introduce inconsistent asymmetric equals behavior.

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/AdaptationHints.java
```diff
@@ -14,20 +15,21 @@
...
+	public static AdaptationHints generateHints(Hints hints, boolean generateID) {
 		// collect all old hints that are not Adaptable
 		Map<String, String> map = new HashMap<>();
 		hints.getHints().entrySet().parallelStream().filter(e -> e
 			.getKey() != Adaptation.prefix).forEach(e -> map.put(e.getKey(), e
 				.getValue()));
 
 		// add Adaptation.NO
-		AdaptationHints newHints = new AdaptationHints(map);
+		UUID id = generateID ? UUID.randomUUID() : hints.executionChainID();
```
ctrueden, line 31: Repetition of `UUID.randomUUID()`. Not the worst thing, because it's a very short function call. But still a possible bug if we change how default UUIDs are generated later.

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/DefaultHints.java
```diff
@@ -15,16 +17,23 @@
 public class DefaultHints extends AbstractHints {
 
 	public DefaultHints() {
-		super(new String[] {Simplification.ALLOWED, Adaptation.ALLOWED});
+		super(new String[] {Simplification.ALLOWED, Adaptation.ALLOWED, DependencyMatching.NOT_IN_PROGRESS});
```
ctrueden, line 20: We should not bake in specific hint pseudo-enum values here. Instead, the default hints should be empty. And e.g. simplification logic should perform the simplification step _unless_ there is a `Simplification.FORBIDDEN` present in the hints. The `Simplification` pseudo-enum should move into the simplification subsystem's package, and be named `SimplificationHints` for clarity. Analogously, `Adaptation` should move into the adapt subsystem package, and `DependencyMatching` into the correct subsystem package. The default behaviors should not have explicit hints. The `DefaultHints` object here should be empty of explicit hint strings.

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/ImmutableHints.java
```diff
@@ -23,14 +24,25 @@ public ImmutableHints(Map<String, String> h) {
 		super(h);
 	}
 
+	public ImmutableHints(Map<String, String> h, UUID historyHash) {
```
ctrueden, line 27: Ideally, the constructor above should be able to call `super` or `this` for DRYness. This could be accomplished by rejiggering the abstract class not to call `setHint`. Really, there are two layers: a protected `populateHint`/`populateHints` style thing, which always works, and a `setHint`/`setHints` public thing which only works if it's allowed, but when it is allowed can delegate to the populate routine, for DRYness.

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/SimplificationHints.java
```diff
@@ -3,40 +3,46 @@
...
+		SimplificationHints newHints = new SimplificationHints(id, map);
 
 		return newHints;
 	}
 
 	@Override
 	public String setHint(String hint) {
-		if (hint.equals(Simplification.ALLOWED)) throw new IllegalArgumentException(
-			"We cannot allow simplification during simplification; this would cause a recursive loop!");
+		if (hint.equals(Simplification.ALLOWED))
```
ctrueden, line 34: Consider whether we need a `SimplificationHints` subclass of the `Hints` interface. Would be nice if not, but I'm naive here. Also, this code would be affected if we switched away from explicit ALLOWED/FORBIDDEN pairs, in favor of sensible default style hints.

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
```diff
@@ -255,95 +247,112 @@ public void register(final OpInfo info, final String name) {
...
+		return (T) wrapViaCache(conditions, conditions.hints().executionChainID());
+	}
+
+	@SuppressWarnings("unchecked")
+	private <T> T findOp(final OpInfo info, final Nil<T> specialType, final Nil<?>[] inTypes,
+		final Nil<?> outType, Hints hints) throws OpMatchingException
+	{
+		OpRef ref = OpRef.fromTypes(specialType.getType(), outType.getType(),
+			toTypes(inTypes));
+		MatchingConditions conditions = MatchingConditions.from(ref, hints, true);
```
ctrueden, line 265: The next three lines are the same as previous method; should be combined into a helper.

I'm also hung up on the name `generateOpInstance`. It _does_ do that :smile: but also puts the result into the cache, and it also returns void. So... how about `ensureCacheHit` or `updateCache` or similar? Personally I like the name `match`. It's clear that this is an important method; it's private, so not too confusing with the actual `OpMatcher` subsystem, and it's accurate. You could also have multiple sigs, one that takes the `MatchingConditions` and one that takes the `OpRef`/`Hints`/`boolean`.

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
```diff
@@ -255,95 +247,112 @@ public void register(final OpInfo info, final String name) {
...
 	 * 
-	 * @param ref
+	 * @param conditions the {@link MatchingConditions} containing the
+	 *          requirements for the returned Op
 	 * @param info
-	 * @return an Op created from {@code info}
 	 * @throws OpMatchingException
 	 */
-	private Object findOpInstance(final OpRef ref, final OpInfo info, Hints hints) throws OpMatchingException {
+	private void generateOpInstance(final MatchingConditions conditions, final OpInfo info) throws OpMatchingException {
```
ctrueden, line 285: Consider rolling in `MatchingConditions.from(ref, hints, true);` here or in another signature, to simplify calling code.

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
```diff
@@ -255,95 +247,112 @@ public void register(final OpInfo info, final String name) {
 	}
 
 	@SuppressWarnings("unchecked")
-	private <T> T findOpInstance(final String opName, final Nil<T> specialType, final Nil<?>[] inTypes,
+	private <T> T findOp(final String opName, final Nil<T> specialType, final Nil<?>[] inTypes,
```
ctrueden, line 250: When reading the source in recent past, I was confused by several `op` versus `findOp` methods. I know some are private, but perhaps there are ways of clarifying the naming here...

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
```diff
@@ -409,7 +428,9 @@ private Object instantiateOp(final OpCandidate candidate, Hints hints)
 	 *            
 	 * @return an {@link Op} wrapping of op.
 	 */
-	private <T> T wrapOp(T op, OpInfo opInfo, Map<TypeVariable<?>, Type> typeVarAssigns) {
+	@SuppressWarnings("unchecked")
+	private <T> T wrapOp(T op, OpInfo opInfo, Hints hints, UUID executionID, Map<TypeVariable<?>, Type> typeVarAssigns) {
+		// TODO: synchronize this
```
ctrueden, line 433: The initWrappers method should be `synchronized`, and should also return immediately if `wrappers != null`. This is double-checked locking, which purists will punch me for suggesting, but it works in practice extremely well.

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
```diff
@@ -464,7 +486,7 @@ private Object instantiateOp(final OpCandidate candidate, Hints hints)
 		return list;
 	}
 
-	private List<Object> resolveOpDependencies(OpCandidate candidate, Hints hints) throws OpMatchingException {
+	private List<MatchingConditions> resolveOpDependencies(OpCandidate candidate, Hints hints) throws OpMatchingException {
```
ctrueden, line 489: :heart:

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
```diff
@@ -488,20 +510,25 @@ private void initWrappers() {
...
 		final List<OpDependencyMember<?>> dependencies = info.dependencies();
-		final List<Object> resolvedDependencies = new ArrayList<>(dependencies.size());
+		final List<MatchingConditions> resolvedDependencies = new ArrayList<>(dependencies.size());
 
 		for (final OpDependencyMember<?> dependency : dependencies) {
 			final OpRef dependencyRef = inferOpRef(dependency, typeVarAssigns);
 			try {
-				Hints hintCopy = hints.getCopy();
+				// TODO: Consider a new Hint implementation
+				Hints hintCopy = hints.getCopy(false);
```
ctrueden, line 522: As above, this is not a getter. So should be called `copy`. But maybe we want two methods? `copy()` and `copyRandomUUID()`?

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
```diff
@@ -488,20 +510,25 @@ private void initWrappers() {
...
+				hintCopy.setHint(DependencyMatching.IN_PROGRESS);
 				hintCopy.setHint(Simplification.FORBIDDEN);
 				if(!dependency.isAdaptable()) {
 					hintCopy.setHint(Adaptation.FORBIDDEN);
 				}
-				resolvedDependencies.add(findOpInstance(dependencyRef, hintCopy));
+
+				MatchingConditions conditions = MatchingConditions.from(dependencyRef, hintCopy, false);
+				generateOpInstance(conditions);
```
ctrueden, line 530: DRYer potential here, depending on changes to `generateOpInstance` :arrow_right: `match` or similar change.

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/LambdaTypeBaker.java
```diff
@@ -0,0 +1,220 @@
...
+import javassist.CtClass;
+import javassist.CtConstructor;
+import javassist.CtField;
+import javassist.CtMethod;
+import javassist.CtNewConstructor;
+import javassist.CtNewMethod;
+import javassist.NotFoundException;
+
+public class LambdaTypeBaker {
+
```
ctrueden, line 23: Should probably make class `final` and add a private ctor, no?

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpMethodInfo.java
```diff
@@ -323,7 +323,7 @@ private String getParameterName(Class<?> c) {
 		// Object[] to ObjectArr. This is not truly extensible, since someone
 		// could create an ObjectArr class which might conflict, so it would be
 		// best to find a better solution.
-		return  c.getComponentType() + "Arr";
+		return  c.getComponentType().getSimpleName() + "Arr";
```
ctrueden, line 326: Use same logic as `LambdaTypeBaker`: `className = className.replaceAll("[^a-zA-Z0-9_]", "_");`

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpWrappers.java
```diff
@@ -26,12 +32,21 @@
 	public static class ProducerOpWrapper<T> implements OpWrapper<Producer<T>> {
 
 		@Override
-		public Producer<T> wrap(final Producer<T> op, final Type reifiedType) {
+		public Producer<T> wrap(final Producer<T> op, final OpInfo info, final Hints hints, final OpHistory history, final UUID executionID, final Type reifiedType) {
```
ctrueden, line 35: Now that `Producer`'s `wrap` signature is longer, we should put the args onto separate lines like below.

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpWrappers.java
```diff
@@ -64,7 +83,15 @@ public Type getType() {
 				@Override
 				public O apply(I in) //
 				{
-					return op.apply(in);
+					// Call the op
+					O out = op.apply(in);
+
+					// Log a new execution
+					if (!hints.containsHint(DependencyMatching.IN_PROGRESS)) {
```
ctrueden, line 90: Let's have a hint namespace for the history itself, with a value like `SKIP_HISTORY` or `SKIP_RECORDING` or some such. Then this gives more explicit control for people who don't want an op execution to be recorded. (Which package?)

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpWrappers.java
```diff
@@ -95,7 +126,15 @@ public Type getType() {
 				@Override
 				public O apply(I1 in1, I2 in2) //
 				{
-					return op.apply(in1, in2);
+					// Call the op
+					O out = op.apply(in1, in2);
+
+					// Log a new execution
+					if (!hints.containsHint(DependencyMatching.IN_PROGRESS)) {
```
ctrueden, line 133: We are discussing a new interface, code-named `RichOp extends GenericTyped`, which all these wrappers extend. It offers accessors for `OpInfo`, `Hints`, `OpHistory`, `UUID`. It has `preprocess(Object... in)` and `postprocess(Object out)` methods. The default `RichOp` implementation would take care of data provenance, progress reporting (in a future PR), and other future bookkeeping things (caching?) would be easily addable later in a uniform way.

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpWrappers.java
```diff
@@ -26,12 +32,21 @@
...
 		@Override
-		public Producer<T> wrap(final Producer<T> op, final Type reifiedType) {
+		public Producer<T> wrap(final Producer<T> op, final OpInfo info, final Hints hints, final OpHistory history, final UUID executionID, final Type reifiedType) {
 			class GenericTypedProducer implements Producer<T>, GenericTyped {
 
 				@Override
 				public T create() {
-					return op.create();
+
+					// Call the op
```
ctrueden, line 41: Tangential to data provenance—more about progress reporting, but also metadata in general—let's try having a static helper `OpMetadata.get()` or some such, which is backed by a `ThreadLocal`. Here, before calling `op.create()` (or whatever functional interface method, we'd write `OpMetadata.set(this)`. Then immediately following `op.create()` we'd write `OpMetadata.set(null)` or `OpMetadata.clear()` or some such.

```java
public final class OpMetadata {
  private OpMetadata() { }
  private static ThreadLocal<RichOp> metadata = new ThreadLocal<>();
  public static RichOp get() { return metadata.get(); }
  public static void remove() { metadata.remove(); }
  public static void set(RichOp opWrapper) { metadata.set(opWrapper); }
}
```

Then in op implementations, even in a lambda for example, you can do:
```
Function<String[], double[]> parseDoubles = s -> {
  RichOp metadata = OpMetadata.get();
  double[] d = new double[s.length];
  ProgressReporter p = metadata.progressReporter();
  p.setProgress(0);
  for (int i=0; i<s.length; i++) {
    d[i] = Double.parseDouble(s[i]);
    p.setProgress((double) i / s.length);
  }
  return d;
}
```

A question is: for registering listeners to be notified about progress, how does that work? Probably a method of `OpEnvironment`:
```java
void addProgressListener(Object op, ProgressListener listener);
```
And then:
```java
public interface ProgressListener {
  void progressUpdated(ProgressEvent e);
}
public class ProgressEvent {
  private RichOp op;
  private double progress;
  public ProgressEvent(RichOp op, double progress) {
    this.op = op;
    this.progress = progress;
  }
  public RichOp op() { return op; }
  public double progress() { return progress; }
}
```
And in `DefaultOpEnvironment`:
```
@Override
private Map<RichOp, List<ProgressListener>> listeners = new HashMap<>();
public synchronized void addProgressListener(RichOp op, ProgressListener l) {
  getListeners(op).add(l);
}
public synchronized boolean removeProgressListener(RichOp op, ProgressListener l) {
  return getListeners(op).remove(l);
}
public synchronized List<ProgressListener> getListeners(RichOp op) {
  // TODO: consider whether to wrap as immutable here, and have a helper getListenersNoWrap method...
  return listeners.getOrDefault(op, ArrayList::new);
}
```
Another TODO is: do we want to also allow registering listeners for _all_ op executions? I.e. for `null` `RichOp`s? This would complicate things, but also be doable as the null key in the listener map.

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpExecutionSummary.java
```diff
@@ -0,0 +1,113 @@
...
+
+import org.scijava.ops.OpInfo;
+import org.scijava.ops.impl.DefaultOpEnvironment;
+
+/**
+ * Describes the execution of an Op
+ *
+ * @author Gabe Selzer
+ */
+public class OpExecutionSummary {
```
ctrueden, line 14: We talked about requirements. The main requirement in my view is: give me a string (that might be a multi-line string) that unambiguously identifies the op implementation(s) and version(s) used, _including dependencies_, for a particular execution. With this string, but the original same inputs, one could then reliably reproduce the same output on any system, even if the environment differs, as long as the implementations at those versions are available in the environment. (And maybe someday, we could even have a layer that dynamically pulls down the needed dependencies at those versions to _ensure_ they are!)

With this in mind, the following ramifications emerge:

1. We should not record the inputs or outputs of the execution. That's up to the person consuming the history to do.
2. We should not keep any references to op instances. This could introduce memory leaks, and it's not necessary to fulfill the requirement described above.
3. Every op implementation+version needs to have a unique string identifier, so that from a given history record, one can lookup the desired ops to use them. E.g. an `OpClass` implementation at version 1.0.0 could be `net.imagej.ops2.vesselness.Frangi@1.0.0` or some such. This can be provided by a new `id()` method of `OpInfo`.

(As a very stupid aside: if an op implementation uses the forthcoming `OpMetadata.get()` and then computes contingent on its `OpInstance`, which knows its `TypeVarAssigns`... then reproducibility would be broken. We should document, in `OpMetadata.get()`, that the op must not compute contingent on such things—it's intended for things like progress reporting, that do not affect the final answer.)

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpExecutionSummary.java
```diff
@@ -0,0 +1,113 @@
...
+	private final UUID executionHash;
+
+	/** The {@link OpInfo} responsible for creating {@code instance} */
+	private final OpInfo info;
+
+	/** The {@link Object} created by {@code info} */
+	private final Object instance;
+
+	/** The {@link Object} returned by the call to {@link DefaultOpEnvironment} */
+	private final Object wrappedInstance;
```
ctrueden, line 26: In line with the above, the cons might outweigh the pros when it comes to keeping hard references to previously-executed ops. Keeping the `OpInfo` is reasonable for sure, because those persist throughout the environment's lifetime. But if we keep references to op instances, then none of them will ever be garbage collected as long as they are in the op history. But... we already cache instances in the op cache, so that they can be reused... so the `instance` here may actually be totally fine. But `wrappedInstance` should probably not be kept, to avoid proliferating wrapped objects forever on the heap.

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistory.java
```diff
@@ -0,0 +1,110 @@
...
+	// -- USER API -- //
+
+	/**
+	 * Returns the list of executions on {@link Object} {@code o} recorded in the
+	 * history
+	 * 
+	 * @param o the {@link Object} of interest
+	 * @return a {@link List} of all executions upon {@code o}
+	 */
+	public List<OpExecutionSummary> executionsUpon(Object o);
```
ctrueden, line 42: `public` is redundant on interfaces.

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistory.java
```diff
@@ -0,0 +1,110 @@
...
+
+import org.scijava.ops.OpInfo;
+import org.scijava.service.SciJavaService;
+
+/**
+ * Log describing each execution of an Op. This class is designed to answer two
+ * questions:
+ * <ol>
+ * <li>Given an {@link Object} output (e.g. a {@code List<String>}), what Op(s)
+ * mutated that output?
```
ctrueden, line 17: More generally: "what Op(s) produced and/or modified this output"?

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistory.java
```diff
@@ -0,0 +1,110 @@
...
+ * <li>Given an {@link Object} op, what {@link OpInfo}s were utilized to
+ * implement that Op's functionality?
```
ctrueden, line 19: I.e. what were the resolved dependencies of a particular op execution?

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistory.java
ctrueden: To achieve `executesUpon(Object out)` without retaining the out reference anywhere in the history explicitly, we can have a `WeakHashMap` that goes from outputs to provenance objects. This is Java's equivalent of "attaching" information to an object without changing its type, in a way that does not block garbage collection.

One weirdness here, to me, is the same object (or equal objects, really) being returned by multiple executions. Now you ask for the executions upon it, and are given a list of length 2, and what do you even do with that? :shrug: But... it's the best we can do. In the vast majority of cases, there will be unique outputs, and only one execution per.

ctrueden: To obtain version of a given op, SciJava Common 2 has a method `VersionUtils.getVersion(Class)`. For us, we will need to either: A) have a `String version()` method of the `OpInfo` interface, which is backed by that; or B) `String id()` of `OpInfo` already bakes in the version with `@` separator. I lean toward (B), but (A) would make it easier to reason that two ops are the same implementation at different versions, without doing string parsing.

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistoryService.java
```diff
@@ -0,0 +1,10 @@
+package org.scijava.ops.provenance;
+
+import org.scijava.service.SciJavaService;
+
+
+public interface OpHistoryService extends SciJavaService {
```
ctrueden, line 6: You could have `OpService` interface possess both an `OpEnvironment` _and_ `OpHistory`. This would simplify the landscape of SciJava services for Ops.

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/impl/DefaultOpHistory.java
```diff
@@ -0,0 +1,241 @@
...
+	// -- HISTORY MAINTENANCE API -- //
+
+	/**
+	 * Logs a {@link OpExecutionSummary} in the history
+	 * 
+	 * @param e the {@link OpExecutionSummary}
+	 * @return true iff {@code e} was successfully logged
+	 */
+	@Override
+	public boolean addExecution(OpExecutionSummary e) {
```
ctrueden, line 139: We discussed how simplification and adaptation should generate IDs for their `OpInfo`s such that they can be reconstructed from those IDs later. Therefore, we need to make sure we test the history functionality with simplification, adaptation, and any other cases relating to ID generation and reexecution.

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/impl/SingletonOpHistoryService.java
```diff
@@ -0,0 +1,25 @@
...
+import org.scijava.service.AbstractService;
+
+
+/**
+ * {@link OpHistoryService} containing a singleton {@link DefaultOpHistory}
+ * 
+ * @author Gabriel Selzer
+ */
+@Plugin(type = OpHistoryService.class)
+public class SingletonOpHistoryService extends AbstractService implements
```
ctrueden, line 15: I prefer `Default` prefix when there is only one. But it's moot if we combine with `DefaultOpService`.

scijava/scijava-ops/src/main/java/org/scijava/ops/simplify/SimplifiedOpRef.java
```diff
@@ -104,7 +106,7 @@ public static SimplifiedOpRef simplificationOf(OpEnvironment env, OpRef ref, Hin
 	 * @throws OpMatchingException
 	 */
 	private static Computers.Arity1<?, ?> simplifierCopyOp(OpEnvironment env, Type copyType, Hints hints) throws OpMatchingException{
-			Hints hintsCopy = hints.getCopy();
+			Hints hintsCopy = new DefaultHints(hints.getHints());
```
ctrueden, line 109: How is this different than `hints.getCopy(true)`? Just a holdover from earlier iterations?

scijava/scijava-ops/src/main/java/org/scijava/ops/util/OpWrapper.java
```diff
@@ -2,12 +2,16 @@
...
 
+import org.scijava.ops.OpInfo;
+import org.scijava.ops.hints.Hints;
+import org.scijava.ops.provenance.OpHistory;
 import org.scijava.plugin.SciJavaPlugin;
 import org.scijava.types.Types;
 
 public interface OpWrapper<T> extends SciJavaPlugin {
-	T wrap(T op, Type reifiedType);
+	T wrap(T op, OpInfo info, Hints hints, OpHistory history, UUID uuid, Type reifiedType);
```
ctrueden, line 14: What we could do is make `RichOp` consist of two things: 1) an `Object op` and 2) `OpBookkeeping stuff`, and `OpBookkeeping` is a container for `OpInfo`, `Hints`, `OpHistory`, `UUID`, and anything else we want to add later.

ctrueden (commented):
:rocket: :guitar: :trophy: But change everything. Just kidding, change some things. :wink: 

For others: we are planning to merge this as is, but later, after reviewing the other PRs in the chain. They will all merge as is (barring major issues), but then we'll have another iteration of work on top of it all, based on these reviews. :+1:

scijava/scijava-ops/src/main/java/org/scijava/ops/OpEnvironment.java
gselzer: Fixed in 6d2a8768

gselzer: TODO: Add methods to `BasicHints` (maybe a `setImmutableHint(String hintType, Object value)`?) to replace most/all of our Hints implementations

scijava/scijava-ops/src/main/java/org/scijava/ops/conversionLoss/LossReporterWrapper.java
gselzer: Fixed in f881e474

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/BaseOpHints.java
gselzer: Fixed in 7837a950

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/Hints.java
gselzer: Fixed with 7837a950

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/Hints.java
gselzer: Fixed with  fc2fb7b2

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/AbstractHints.java
gselzer: Fixed by 017f633e

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/AbstractHints.java
gselzer: Fixed by 017f633e

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/AbstractHints.java
gselzer: Fixed with 017f633e

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/AdaptationHints.java
gselzer: Fixed with d142781b

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/DefaultHints.java
gselzer: Hmm, I'm not sure that I like that move. We need to `export` the `Simplification` pseudo-enum, but I'd prefer not to export the package `org.scijava.ops.simplify` (later to become `org.scijava.ops.engine.simplify`). I *suppose* we could rename `org.scijava.ops.engine.simplify` ➡️ `org.scijava.ops.engine.simplify.impl`, then make a new package `org.scijava.ops.engine.simplify` with _just_ `SimplificationHints`? That doesn't seem all too bad to me...

(Of course, we'd do the same thing for Adaptation, and later Reduction)

ctrueden: I thought the `scijava-ops-engine` module was implementation-only, and exports nothing? No? If so, I'd like to keep it that way. Or if not, I'd like to understand better what it's exporting (presumably for compile-time dependencies downstream).

I think the hints pseudo-enums need to go into the `scijava-ops-api` module. Which package is something to discuss—obviously an exported one (does `scijava-ops-api` even have any internal packages?). :smile:

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/ImmutableHints.java
gselzer: Fixed with 0e8aabd4

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/DefaultHints.java
gselzer: > Or if not, I'd like to understand better what it's exporting (presumably for compile-time dependencies downstream).

There are three packages exported:
* `org.scijava.ops.engine`: Contains `OpService` / `OpHistoryService`, which will later go away, and some other API classes that could presumably go into `scijava-ops-api`. And then `BaseOpHints`, which contains the pseudo-enum values
* `org.scijava.ops.engine.conversionLoss`: Contains the `LossReporter` interface, which is an implementation detail of the Simplification framework
* `org.scijava.ops.engine.util`: Contains utility methods useful to Op implementations. I'd like to delete this package at some point.

> I think the hints pseudo-enums need to go into the `scijava-ops-api` module.

I disagree. They all pertain to matching, which is restricted to the `scijava-ops-engine` module. `scijava-ops-api` should not concern itself with these implementation details.

> (does `scijava-ops-api` even have any internal packages?).

Nope :grin:

ctrueden: What about merging `OpService` and `OpHistoryService`, and moving them to `scijava-ops-services`? This would be a step forward toward eliminating the scijava-common dependency from `scijava-ops-engine`. (It is a necessary but not sufficient step.) Not necessarily now, but let's discuss further after the PR chain is complete.

gselzer: @ctrueden do 1f0d56a3 and fb5e5037, combined with scijava/scijava#71 and scijava/scijava#72 cover everything we discussed here?

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistoryService.java
gselzer: fixed in 5a76dcd5

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/impl/SingletonOpHistoryService.java
gselzer: Fixed in 5a76dcd5

scijava/scijava-ops/src/main/java/org/scijava/ops/simplify/SimplifiedOpRef.java
gselzer: For some reason, changing this back to `hints.getCopy(true)` creates test failures. Will investigate.

gselzer: Explained in b13c9964, should be fixed in scijava/scijava#72

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/DefaultHints.java
ctrueden: Yep, looks good! The only other related thing we discussed, which should perhaps be another issue, is the idea that `scijava-ops-api` contains more than one layer of abstractions, which could conceivably be split into two modules later if we decide it would be nicer: 1) the base layer of interfaces for people who want to just use the `OpEnvironment` to execute ops, and the more specialized interfaces pertaining to features the majority of users won't care about explicitly: adaptation, simplification, loss reporting, etc. My preference, for now, would be to delineate these more specialized interfaces into one or more subpackages of `.api`, so that it is clear they are less central than then things in `.api` directly. What do you think? For the moment we can just file an issue.

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/impl/DefaultOpHistory.java
gselzer: IDs added with 9fcbc9e8

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/DefaultHints.java
gselzer: `BaseOpHints`have been moved to package `org.scijava.ops.api.features` with 4c215f59; not satisfied with the naming, though. Relevant issue about module separation filed as scijava/scijava#73

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
gselzer: Fixed in de4a2c18

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/LambdaTypeBaker.java
gselzer: Fixed with 8cde9ff0

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpMethodInfo.java
gselzer: Fixed in 9005ce59

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpWrappers.java
gselzer: Fixed in 0f02e169

scijava/scijava-ops/src/main/java/org/scijava/ops/util/OpWrapper.java
gselzer: @ctrueden this change is made in 8a87442e.

We now have `RichOp`, `DefaultRichOp`, and `OpMetadata`. Part of me wonders if we actually need all of these types, but I think they are nice.

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpExecutionSummary.java
gselzer: Changes made in 5e04129f. 

My one concern with that commit is that `DefaultOpHistory.addExecution` (as of this writing) is unable to cache different `Object`s with the same hashcode. This change actually required an alteration to `ProvenanceTest`, as we _were_ able to differentiate two `Double`s of the _same value_ returned by two different `Function`s. 

Specifically, we had `Double d1` returned by `Function f1`, and `Double d2` returned by `Function f2` (where `d1.doubleValue() == d2.doubleValue()` but `d1 != d2`), where `OpHistory.executionsUpon(d1)` would point to _only_ `f1` and `OpHistory.executionsUpon(d2)` would point to _only_ `f2`. Now, we have `OpHistory.executionsUpon(d1)` pointing _both_ `d1` and `d2` (since `d1.hashCode() == d2.hashCode()`, which is wrong. 

Don't have a good idea of fixing this though, without adding complexity to the implementation 🙁

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistory.java
gselzer: Fixed with b09dc793

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistory.java
gselzer: Fixed with b09dc793

scijava/scijava-ops/src/main/java/org/scijava/ops/matcher/OpWrappers.java
gselzer: Fixed in 48ce1a05

scijava/scijava-ops/src/main/java/org/scijava/ops/provenance/OpHistory.java
gselzer: The first comment was fixed by 7e600df6 
The second was fixed by 5e04129f 
The third was fixed by 09eeca92

scijava/scijava-ops/src/main/java/org/scijava/ops/OpEnvironment.java
gselzer: Fixed with fbaeef1e

scijava/scijava-ops/src/main/java/org/scijava/ops/hints/impl/SimplificationHints.java
gselzer: Fixed with fbaeef1e

scijava/scijava-ops/src/main/java/org/scijava/ops/impl/DefaultOpEnvironment.java
gselzer: Fixed in 48ce1a05bf6181dc8b2ad83932d710f30c952ad2

PR #36, opened 2021-06-24
RealTransform

gselzer:
This PR is a port of https://github.com/imagej/imagej-ops/pull/533, thanks to @bnorthan for the initial work.

Note that in an ideal world we would be able to port the optionality of the later parameters of this Op, however that functionality does not quite exist. The functionality should be added once #32 is merged.


gselzer:
@ctrueden @hinerm I updated this PR based on the advancements we've made to the project over...the last couple years :sweat_smile:.

I think it's ready for merge, unless we're concerned about the test resources added - I think it's difficult to do a mathematical regression test, though, due to the rounding all over the place, and the resources are pretty small, so I'm not too concerned.

ctrueden (approved):
Thanks @gselzer! The binary test resources are no problem, as they are lightweight.

PR #37, opened 2021-07-15
Prioritize a more powerful version of setHints

gselzer:
Previously we prioritized a setHints taking only the whole Hint. Now we
prioritize a Hint composed of the String hintType and the Object value.
We also add a getDelimiter method to provide an
implementation-independent way to get the delimiter. In this way we no
longer tie ourselves to a single delimiter.

Inspired by a conversation with @ctrueden on Gitter

The singular commit builds with passing tests


ctrueden:
After b4169e962afffe962c86d04817c1cbeb770b4acc, this PR might be obsolete. Or it might have some goodness to be extracted. Scrutiny needed.

gselzer:
Looking through the diff, I can't find anything that I think needs to be ported!

PR #38, opened 2021-07-28
Partition SciJava Ops into separate modules

### LINKED ISSUE #63: Remove/refactor packages outside org.scijava.ops
We currently have the packages `org.scijava.param` and `org.scijava.struct`, which **should not** remain in this project. The code should either be moved into its own module, or the packages refactored. 

We should furthermore enforce via a Junit test that **all** packages within the `org.scijava.ops` module have the prefix `org.scijava.ops` . This test should live within [`ModuleTest`](https://github.com/scijava/incubator/blob/scijava/scijava-ops/view-javadoc/scijava/scijava-ops/src/test/java/org/scijava/ops/ModuleTest.java) (which is currently on a branch), or something like `ModuleValidityTest` (we could rename `ModuleTest`).

Somewhat related to #62, weakly related to #61

Time commitment: 1 week

gselzer:
This PR partitions SciJava Ops into a set of new JPMS/Maven modules:
* SciJava Ops API: provides a set of external API for use by third party libraries who **only** wish to use SciJava Ops / ImageJ Ops2
* SciJava Ops SPI: provides a set of annotations/service interfaces for use by third party libraries who wish to **extend** SciJava Ops / ImageJ Ops2 with **their own implementations**
* SciJava Ops Engine: provides a default Op matching implementation, with Simplification, Adaptation, etc.
* SciJava Ops Discovery: provides an interface for class discovery. Currently, SciJava Ops Engine implements this with Plugin Discovery, but we could easily add a Spring Boot implementation

This allows much more control over which sections of the library will be exported, and will result in cleaner code. This will also narrow our dependency on various libraries to certain submodules of SciJava Ops, making it easier to migrate the library to new implementations of, say, dependency injection or implementation discovery.

TODO:
* [ ] Where should we store the API tests (for example, `OpBuilderTest`)? Ideally, we'd keep them within SciJava Ops API, however we don't have the implementations required to thoroughly test, for example, `OpBuilder` within SciJava Ops API. We could always keep them within SciJava Ops Engine, but this seems like bad practice. \
* [ ] Better partition the SciJava Common dependency. This is hard, because all Ops (both test and main Ops) currently have the dependency on `Plugin`.
* [ ] Merge #28, #31, #39

Closes scijava/scijava#63, scijava/scijava#61


ctrueden:
> Where should we store the API tests

I don't know that there's really a such thing as an "API test"—only tests of implementations.

That said, now that Java interfaces can have default implementations, it can get tricky to test those default methods in the API component while testing implementation(s) of abstract methods in the components that implement them. My vote would be to just keep all the tests in the engine component, which is where the implementation gets completed.

gselzer:
> > Where should we store the API tests
> 
> I don't know that there's really a such thing as an "API test"—only tests of implementations.
> 
> That said, now that Java interfaces can have default implementations, it can get tricky to test those default methods in the API component while testing implementation(s) of abstract methods in the components that implement them. My vote would be to just keep all the tests in the engine component, which is where the implementation gets completed.

Yeah, it just sucks that we have to have the `OpBuilder` implementation in API. We *could* make an interface for `OpBuilder`...

gselzer:
This would also close scijava/scijava#59

PR #39, opened 2021-07-28
Enforce package hierarchy

### LINKED ISSUE #62: Enforce better package relationships at build time
__No circular dependence between packages.__ E.g. if `org.scijava.ops.a` imports something from `org.scijava.ops.b`, then `org.scijava.ops.b` must not also import something from `org.scijava.ops.a`—_including subpackages_.

__No dependence on subpackages.__ E.g. Classes of `org.scijava.ops` must not import classes from `org.scijava.ops.a`, `org.scijava.ops.a.b`, etc.

To facilitate this, we should employ a dependency analysis step during the build, and fail if these rules are broken. We could use [JDeps](https://docs.oracle.com/en/java/javase/11/tools/jdeps.html) and/or [JDepend](https://github.com/clarkware/jdepend) (see also [jdepend-maven-plugin](https://github.com/mojohaus/jdepend-maven-plugin)).

gselzer:
This PR adds a set of package enforcement rules to all incubator builds. As of this writing, we enforce two rules:
* No package cycles
* No subpackage dependence

TODO:
* [x] merge scijava/scijava-packages-rules#1, and cut a release of that project
* [x] Add a commit updating the version of that artifact
* [x] merge #28, #31

Closes scijava/scijava#62


ctrueden (approved):
I released a new scijava-maven-plugin 2.0.0 with the two new package rules, and updated this PR accordingly, such that there are no more WIP commits. LGTM now.

PR #40, opened 2021-07-29
Write READMES for each of the current incubator projects 

gselzer:
Closes https://github.com/scijava/scijava/issues/60

TODO:
* [x] SciJava Ops API
* [x] SciJava Ops SPI
* [x] SciJava Ops Discovery
* [x] SciJava Ops Engine
* [x] SciJava Types
* [x] SciJava Struct
* [x] SciJava Function
* [x] SciJava Taglets
* [x] SciJava Persist
* [x] SciJava Test Utilities
* [x] ImageJ Test Utilities
* [x] ImageJ Ops 2
* [x] Incubator

Further TODO:
* [x] Merge https://github.com/scijava/incubator/pull/39

PR #41, opened 2021-08-02
ServiceLoader-based Discoverer

gselzer:
This PR introduces a [`ServiceLoader`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html)-based `Discoverer`, currently able to discover `Op`s and `OpCollection`s. Suppose we want make a class `FooOp` in module `com.example.foo` discoverable by `ServiceLoaderDiscoverer`; this can be accomplished by following these steps:

1. Make `Cls` implement `Op` or `OpCollection`:
```java
class Cls implements Op {}
```
2. In `module-info.java`, `provide` `Op` with `Cls`:
```java
module com.example.foo {
  provides org.scijava.ops.spi.Op with Cls;
}
```

This PR also adds `OpClass` to the SPI module, intended to house Op metadata normally stored in a `@Plugin` annotation, as well as an Op implementation in the SciJava Ops Engine module as a proof of concept.

TODO:
* [ ] Consider logging a warning for any class implementing `Op` or `OpCollection` that **is not** declared in the `module-info`. This should be enforced via a Maven plugin
* [x] Merge #40, as this branch is built on top of that work.


gselzer:
# NB: An answer to the question: "Why must this depend on SciJava Ops SPI?:

The big reason is to be able to discover `OpCollection`s. We need a marker interface for the `ServiceLoader` to `use` in order to discover `OpMethod`s or `OpField`s. Since we already need `scijava-ops-spi` for the `OpMethod`/`OpField` annotations, we might as well implement `OpCollection`. But, let's assume that we didn't care about this, and focus only on exposing Classes that we wish to indicate should be Ops.

The `ServiceLoader` paradigm requires us to `provide` some interface to make an implementation exposed. Since Ops must be `FunctionalInterface`s, we then have but two choices:
1) Provide the `FunctionalInterface` (e.g. `Function`)
2) Provide `Op`

One downside of (1) is that we may pick up `Function`s that were not intended to become Ops. Of course, since Ops need to have the `OpClass` annotation anyways, we could filter out any `Function`s that don't have annotations, so this part could be avoided. The other downside is that this is not extensible: for the `ServiceProvider` to discover *any* `FunctionalInterface`s we would have to modify this `module-info`, and thus introduce a runtime dependency on `scijava-function` or whatever other dependency has the interface. This is not ideal

For these reasons, it is simplest to just use `Op` and `OpCollection`. Thoughts?

ctrueden (approved):
This was a great first step toward using `ServiceLoader` as an option for discovery of ops. We have discussed and iterated the design several times since then, but the latter work is built on this first iteration, so I am merging this as is, with the understanding that the `ServiceLoader`-based discovery evolves substantially from this point forward.

PR #42, opened 2021-09-16
Rewrite DefaultOpMatcher as a set of MatchingRoutines

gselzer:
This PR adds the `MatchingRoutine` interface, an abstraction designed to increase the extensibility of the `DefaultOpMatcher`. Previously, `DefaultOpMatcher` consisted of a hardcoded ordering of matching routines: first a direct (runtime-safe) match was searched for, then an adapted match, and finally a simplified match. Other matching routines, such as the reduction routine of #32 would have to be explicitly added to this chain, or be shoehorned into one of the existing routines (the reduction routine was shoehorned into the direct routine by expanding the number of initial `OpInfo`s).

By adding the `MatchingRoutine` interface, altering the order of these matching routines becomes much easier, and the routines more isolated.

TODO:
* [x] Merge #41 
* [x] Consider using another discovery mechanism for the discovery of the various `MatchingRoutine`s. They are currently supported by the `PluginService`, and it would be nice to switch to SciJava Ops ServiceLoader. This can't really be done without putting `MatchingRoutine` in SciJava Ops SPI, however, and I am not sure whether this is something we want to do.


ctrueden (approved):
I know that later work transitions the `MatchingRoutine`s (and other plugin types) away from SJC2 in favor of JPMS service loader discovery, so I think this branch is good to go.

PR #43, opened 2021-09-30
Progress Reporting Implementation

gselzer:
This PR introduces a mechanism for Progress Reporting within the Ops framework. The core of this mechanism is the static `Progress` class, which is used both for Ops to broadcast updates to their progress and for users to be notified of updates to Ops' progress. The `Progress` class maintains a set of `Task`s, describing all concurrently-executing processing hierarchies. Below details the (current) methods by which each party interfaces with the `Progress` class:

## 1: Ops Implementing Progress Reporting
Ops obtained via the matcher are automagically registered as `Task`s, recorded within `Progress`. Therefore, in order to report numerical progress, Ops must:
1. Define the number of stages it will run, and the number of subtasks (using the command `Progress.defineTotalProgress(int numStages, int numSubTasks)`):
a. A stage is defined as a section of Op processing done within the Op itself, and are customizable within the Op
b. A subtask is defined as a standalone task that will report its progress to this Op. Any Op dependency execution is automatically registered as a subtask.
2. For each stage, the Op must provide the number of updates it plans to record. This can be done calling `Progress.setStageMax(long max)`
After this setup, the Op can report its progress by calling `Progress.update()`.
Here is an example of how an Op might define progress:
```java
@Plugin(type = Op.class, name = "create.array")
public class ArrayCreator implements BiFunction<Integer, Integer, Integer[]> {

  public Integer[] apply(Integer size, Integer initialValue) {
    Progress.defineTotalProgress(1, 0); // 1 Op stage, 0 Op Subtasks
    Progress.setStageMax(size); // Set the progress maximum to size
    Integer[] arr = new Integer[size];
    for (int i = 0; i < arr.length; i++) {
      arr[i] = initialValue;
      Progress.update(); //Record update to the total progress
    }
  }
}
```
Ops can also update their status, without any setup, using the method `Progress.setStatus(String status)`:
```java
@Plugin(type = Op.class, name = "create.array")
public class ArrayCreator implements BiFunction<Integer, Integer, Integer[]> {

  public Integer[] apply(Integer size, Integer initialValue) {
    Integer[] arr = new Integer[size];
    for (int i = 0; i < arr.length; i++) {
      arr[i] = initialValue;
      Progress.setStatus("Initialized element " + i); //Update status
    }
  }
}
```
## 2: The `ProgressListener` interface as an interrupt mechanism for progress updates

Progress updates can trigger interrupt code via the `ProgressListener` interface. This interface is a `FunctionalInterface`, with a functional method `acknowledgeUpdate(Task task)` (allowing a `ProgressListener` to be defined as a lambda). `ProgressListeners` can be registered using the method `Progress.addListener(Object progressible, ProgressListener p)`; an example of a `ProgressListener` interrupt is shown below:
```java
BiFunction<Integer, Integer, Integer> op = ops.op("create.array").inTypes(Integer.class, Integer.class).outType(Integer.class).function();
Progress.addListener(op, (task) -> {
  System.out.println("Progress: " + task.progress());
});
op.apply(400, 1);
```
During the op call, the lambda-defined `ProgressListener` will function as a callback Object, with its `acknowledgeUpdate(Task task)` functional method being called every time the Op calls `Progress.update()`. The method `Task.status()` can also be called if the listener is interested in the status of the Op.


gselzer:
Update: `Progress` and related code have been relocated into a new module, `scijava-progress`

gselzer:
Closes scijava/scijava#42

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/AbstractRichOp.java
```diff
@@ -43,14 +44,15 @@ public OpMetadata metadata() {
 	}
 
 	@Override
-	public void preprocess(Object... inputs) {}
+	public void preprocess(Object... inputs) {
```
ctrueden, line 47: Could be a default method of `RichOp` interface.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/AbstractRichOp.java
```diff
@@ -43,14 +44,15 @@ public OpMetadata metadata() {
...
+		Progress.register(this);
+	}
 
 	@Override
 	public void postprocess(Object output) {
 		// Log a new execution
-		if (!metadata.hints().contains(History.SKIP_RECORDING)) {
-			metadata.history().addExecution(this, output);
-		}
+		metadata.history().logOutput(this, output);
```
ctrueden, line 54: If you change this to `metadata().history()...`, this could be a default method of `RichOp` interface.

scijava/scijava-progress/LICENSE.txt
```diff
@@ -0,0 +1,24 @@
+Copyright (c) 2016 - 2019, SciJava Ops developers.
```
ctrueden, line 1: 2019, heh. We'll have to run a full license update on the codebase after all the PRs are through.

scijava/scijava-progress/src/test/java/org/scijava/progress/ImproperReportingTest.java
```diff
@@ -0,0 +1,156 @@
...
+		Progress.complete();
+	}
+
+	/**
+	 * Tests that tasks who updated progress past the defined maximum result in a
+	 * thrown error.
+	 */
+	@Test
+	public void testDefineTooFewStages() {
+		Supplier<Integer> task = tooFewStageTask;
```
ctrueden, line 106: Why is this supplier external? It's only referenced here once, no? Can we use a lambda?

ctrueden: And same/analogous question for the other functional tasks that are externally defined but could be inlined.

ctrueden (approved):
I had a lot of problems trying to test and review this code from Eclipse. I cleaned up many of the problems and merged the work to main, then rebased this branch over it, but there will continue to be problems until #57 is merged due to Eclipse's failure surrounding annotation processing. Even with the EclipseHelper. I played with augmenting the Ruby script that merges main and test annotations, to make it also clean up after Eclipse, but couldn't make it work out of the box no matter what I tried. Even with building from the command line via `mvn clean install` followed by a refresh in Eclipse, the unit tests in Eclipse _still_ are crazy and can't find plugins and a bunch of them fail. So I'm giving up: I'm just going to push through all these PRs as fast as I can until after #57. Then I hope it will become possible to develop this project in Eclipse without so much pain.

scijava/scijava-progress/LICENSE.txt
gselzer: :sweat_smile: :older_man:

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/AbstractRichOp.java
gselzer: @ctrueden idk, I'm not a huge fan. The default behavior across the incubator is "by default, do nothing".

scijava/scijava-progress/src/test/java/org/scijava/progress/ImproperReportingTest.java
gselzer: I suppose we don't need the variable, we could just pass the supplier in the `Assertions.assertThrows` call below. Or are you suggesting we inline the lambda itself? I think that the latter would be a bad idea.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/AbstractRichOp.java
ctrueden: > "by default, do nothing"

By which you mean: "interfaces do not have defaults"?

In my view:
```
default void preprocess(Object... inputs) {}
```
in `RichOp.java` would be precisely "by default, do nothing". So I'm not sure I understand what you mean here.

scijava/scijava-progress/src/test/java/org/scijava/progress/ImproperReportingTest.java
ctrueden: > Or are you suggesting we inline the lambda itself? I think that the latter would be a bad idea.

That is what I'm suggesting. Why would it be a bad idea?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/AbstractRichOp.java
gselzer: Ah, I thought you were suggesting we push that functionality upstream.

Fixed in 76176433.

scijava/scijava-progress/src/test/java/org/scijava/progress/ImproperReportingTest.java
gselzer: My main concern was readability.

I inlined them in 307d0f39 though.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/AbstractRichOp.java
gselzer: Eh...by default I'd prefer to do nothing in `RichOp`. Do you strongly about this @ctrueden?

PR #44, opened 2021-10-18
Therapi-based Discovery

### LINKED ISSUE #77: Create a Therapi-based Discoverer for Ops
Use case: imglib2-algorithm

**Suppose a library's Javadocs are saved using the therapi annotation processor. Can we use the javadoc at runtime together with an `OpTaglet` to determine which algorithms should be wrapped into Ops, without any additional Maven dependencies?**

Imglib2-algorithm algorithms would then need to declare in their javadoc:
```java
/**
	 * Apply Gaussian convolution to source and write the result to output.
	 * In-place operation (source==target) is supported.
	 * 
	 * ...
	 * 
	 *  @implNote op gauss org.scijava.function.Computers.Arity1
	 * 
	 * @param sigma
	 *            standard deviation of isotropic Gaussian.
	 * @param source
	 *            source image, must be sufficiently padded (e.g.
	 *            {@link Views#extendMirrorSingle(RandomAccessibleInterval)}) to
	 *            provide values for the target interval plus a border of half
	 *            the kernel size.
	 * @param target
	 *            target image
	 * @param <S>
	 *            source type
	 * @param <T>
	 *            target type
	 * @throws IncompatibleTypeException
	 *             if source and target type are not compatible (they must be
	 *             either both {@link RealType RealTypes} or the same type).
	 */
	public static < S extends NumericType< S >, T extends NumericType< T > > void gauss( final double sigma, final RandomAccessible< S > source, final RandomAccessibleInterval< T > target ) throws IncompatibleTypeException
	{
		...
	}
```

In the example, we use the `@implNote` taglet, [introduced with Java 8](https://nipafx.dev/javadoc-tags-apiNote-implSpec-implNote/). Other options include `@apiNote`, `@implSpec`. This taglet will need to declare:
* That it is an op (via the `op` after the taglet)
* The name under which the op will be discovered (`gauss` in the example)
* The functional interface under which the op will be discovered (`org.scijava.function.Computers.Arity1` in this example). Shorthands (such as `Computers.Arity1` should also be supported).

The advantage here is **zero** additional dependencies (although it will require an annotation processor dependency for Ops to discover these taglets and the corresponding algorithm). 

Disadvantages compared to a compile-time dependency on the Ops framework:
* The annotation interfaces (`@OpField`, `@OpMethod`, etc) provide more type safety and discoverability in IDEs. You will not be able to search for imglib2 algorithms as they will not be declared under any annotation. You also cannot search for these ops under, for example, `Function`.
* These annotations can be malformed and the compiler will not recognize the problem.
* This mechanism used was not originally designed for the task. It is unlikely that anyone else will not use this tag, and could conceivably break things (if Java/someone else is tuned into these tags). 
* This requires that, if JPMS modules are used, every package with ops must be `open`ed to both therapi and some SciJava Ops-based module that parses this syntax. **BUT** this is a common issue; **any** component with a `module-info` must `open`/`export` any packages with algorithms designed to be called by other modules.

The annotation processor can be activated by a opt-in property `<therapi.packages>` in the POM. This property can include a comma-delineated set of packages to process only those packages, or left blank to include all packages.

gselzer:
This PR adds a Therapi-based Op Discovery mechanism via the new `scijava/scijava-discovery-therapi` module. This module enables "element-tagged" discovery through the use of the [`@implNote`](https://nipafx.dev/javadoc-tags-apiNote-implSpec-implNote/) javadoc tag.

### How do I declare an Op using Therapi?
1. Enable therapi annotation processing in your POM.
2. Add an `@implNote` to any `Class`, `Method` or `Field`. The `@implNote` annotation should be followed with (1) the tag type (i.e. the keyword under which this `AnnotatedElement` will become discoverable) and (2) any (comma-delineated) key-value pairs that should be scraped with the tag.

*Note that the tags are parsed using Parsington, as the `org.scijava.parse` package from SciJava Common was moved into the incubator and modularized into SciJava Parse2.*

An Op written as a `Field` discoverable using therapi might look like:
```java
/**
 * @implNote op names='foo', priority='100'
 */
public final Function<Integer, Integer> fooOp = (in) -> {...};
```

Ops written as `Method`s **require** a `type` option:
```java
/**
 * @implNote op names='bar', type='java.util.function.Function'
 **/
public static Integer bar(Integer in) { ... }
```

This PR also refactors the `Discoverer` and `Discovery` API from SciJava Discovery to make the discovery returns much more powerful.  

Furthermore, this PR also introduces SciJava Parse2 to the incubator, partitioning out [`org/scijava/parse`](https://github.com/scijava/scijava-common/tree/master/src/main/java/org/scijava/parse) from SciJava Common.

Closes scijava/scijava#77


gselzer:
Note: This is the first step towards completing https://github.com/scijava/scijava/issues/56

scijava/scijava-discovery-therapi/src/test/java/module-info.test
```diff
@@ -0,0 +1,2 @@
+--add-reads
+	org.scijava.discovery.therapi=org.scijava
```
ctrueden, line 2: What is this file? `module-info.test`? I've not heard of this, and cannot find anything with a quick web search. There is no explanation in the corresponding commit about what it is. Is it just cruft that should be deleted?

ctrueden (approved):
:rocket:

PR #45, opened 2021-11-10
Scijava/scijava function/kotlin

elect86:
Code generation via KSP directly in Kotlin, full IDE support and debugging.

Everything nicely integrated with Gradle:
- generation is automatically called before the sources build
- it's also appended to the clean process
- takes advantage of incremental compilations and other optimizations
- no need to do things manually anymore

I worked on Idea and also tested quickly the latest Eclipse and it seems working fine too


ctrueden (changes requested):
I rebased this PR over the latest main branch. Unfortunately, it fails to build the scijava-ops-api component, because the `OpBuilder` code does not get generated.

I suspected it might be my rebase which broke things, so I checked out the original version of the work, c8ab855f6ed92a8865a0ae5bd71f7cfc99a48fc4, and tried building that. It had immediate problems with the maven-enforcer-plugin which we have since solved/avoided, so I applied a small patch to the toplevel pom.xml.

<details><summary>Here is the patch I applied</summary>

```diff
diff --git a/pom.xml b/pom.xml
index 37da3c3f..7c9b6038 100644
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.scijava</groupId>
 		<artifactId>pom-scijava</artifactId>
-		<version>30.0.0</version>
+		<version>33.2.0</version>
 		<relativePath />
 	</parent>
 
@@ -140,41 +140,6 @@
 			</plugin>
 			<plugin>
 				<artifactId>maven-enforcer-plugin</artifactId>
-				<dependencies>
-<!--					<dependency>-->
-<!--						<groupId>org.scijava</groupId>-->
-<!--						<artifactId>scijava-packages-rules</artifactId>-->
-<!--						<version>0-SNAPSHOT</version>-->
-<!--					</dependency>-->
-				</dependencies>
-				<executions>
-					<execution>
-						<id>enforce-no-package-cycles</id>
-						<goals>
-							<goal>enforce</goal>
-						</goals>
-						<phase>test</phase>
-						<configuration>
-							<rules>
-								<NoPackageCyclesRule
-									implementation="org.scijava.packages.rules.NoPackageCyclesRule" />
-							</rules>
-						</configuration>
-					</execution>
-					<execution>
-						<id>enforce-no-subpackage-dependence</id>
-						<goals>
-							<goal>enforce</goal>
-						</goals>
-						<phase>test</phase>
-						<configuration>
-							<rules>
-								<NoSubpackageDependenceRule
-									implementation="org.scijava.packages.rules.NoSubpackageDependenceRule" />
-							</rules>
-						</configuration>
-					</execution>
-				</executions>
 			</plugin>
 		</plugins>
 		<pluginManagement>
```

</details>

Unfortunately, the original code also fails with the same error. Here it is:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project scijava-ops-api: Compilation failure
[ERROR] /home/curtis/code/scijava/incubator/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java:[149,9] error: cannot find symbol
[ERROR]   symbol:   class OpBuilder
[ERROR]   location: interface OpEnvironment
```

I suspect that it never worked correctly on the command line, only perhaps from IDEs. Further investigation is needed to determine how to fix this. Then we can evaluate whether this PR is a direction we would like to go.

My 2c: I am personally very skeptical of Gradle. It is a big step away from standard declarative builds toward ad hoc imperative ones, which I oppose. Furthermore, I am concerned about Gradle's lack of commitment to backward compatibility. The last time I interacted with it (for the [ClearVolume](https://github.com/ClearVolume/clearvolume) project), an existing Gradle <7 build failed to work anymore with Gradle 8+ due to retired plugins: e.g. `maven` became `maven-publish`. Whereas with Maven, I just tried building the ImageJ2 project from a >8-year-old tag, using Maven 3.8.5, and it was still possible thanks to Maven's repository mirrors feature that lets you override where remote artifacts come from in case remote repositories become defunct. Unfortunately, Gradle is not a good choice for scientific projects because open science + source needs reproducibility, but Gradle builds are not long-term reproducible. It would be much better if Gradle would retain support for old builds somehow, e.g. by invoking previous plugin versions which are left continuously available, rather than failing the build.

ctrueden:
OK good news, I was being dumb about the build in that I only tried `mvn` and not `gradle` from the CLI. Using `gradle` (version 7.2), and tweaking one small dependency issue in the `.gradle.kts` files (now squashed in), does result in a successful gradle build. However, it still doesn't work to run `mvn` afterward, continuing to fail with the missing `OpBuilder`. So I'm not sure how this was intended to be used on the CLI. My impression of this PR—but maybe I misunderstand—is that this change is not a complete changeover to gradle as a build system, since `gradle tasks` does not list all the goodies like compile, tests, package, etc. Rather, it's just for the code generation, no? So then... the Maven build needs to be made harmonious with this gradle codegen step... which from the patch here, looks like was the intent...

Anyway, I still think this PR is going to take a lot to digest, to figure out whether this work should be continued...

hinerm:
@elect86 can you explain more about this PR?

>I worked on Idea and also tested quickly the latest Eclipse and it seems working fine too

I take this to mean that the expected behavior is, in Eclipse or Idea, if I modify a `.vm` then the corresponding `.java` file should be immediately regenerated - is that correct? If so, that is not happening for me currently. Is there any additional setup that's required in either IDE?

FWIW in IntelliJ I get this message:
image

What is the intended invocation to build the project with `gradle`? If so, can the `.github` file be updated as well to reflect this? `gradle build` at the top-level didn't work for me; I had to go into individual subdirs, and the build failed with compilation errors:
image

Building with `mvn` has different compilation errors because the autogenerated files aren't being re-generated.

Is there any expected impact to this PR outside of the autogenerated files?

elect86:
Too much divergence and mess, I'll close this and create a new one

ctrueden:
Thanks @elect86, much appreciated! If you can also briefly describe the intended workflow, how to test, etc., that would be super helpful for understanding the benefits this work provides. :+1:

elect86:
> My 2c: I am personally very skeptical of Gradle. It is a big step away from standard declarative builds toward ad hoc imperative ones, which I oppose. Furthermore, I am concerned about Gradle's lack of commitment to backward compatibility. The last time I interacted with it (for the [ClearVolume](https://github.com/ClearVolume/clearvolume) project), an existing Gradle <7 build failed to work anymore with Gradle 8+ due to retired plugins: e.g. maven became maven-publish. Whereas with Maven, I just tried building the ImageJ2 project from a >8-year-old tag, using Maven 3.8.5, and it was still possible thanks to Maven's repository mirrors feature that lets you override where remote artifacts come from in case remote repositories become defunct. Unfortunately, Gradle is not a good choice for scientific projects because open science + source needs reproducibility, but Gradle builds are not long-term reproducible. It would be much better if Gradle would retain support for old builds somehow, e.g. by invoking previous plugin versions which are left continuously available, rather than failing the build.


Gradle is actually even better in terms of reproducibility.

What you are complaining is that a build script written for Gradle 7 doesn't work out-of-the-box with Gradle 8+. 
This is actually a huge advantage of Gradle, if a breaking change makes sense, they do it. Backwards compatibility at all cost is a big error in the long time. Look how held (and it's still partially) back Java development.

Give a look at the development amount just in a [week](https://github.com/gradle/gradle/pulse)
Maven is pretty stale and cannot evolve too much due to the backwards compatibility.
But sometimes it is still the case that a new Maven version cannot run an old Maven build because of incompatible changes.
That's one of the major advantages of the Gradle Wrapper.
Each and every project using Gradle should have the 4 Gradle Wrapper files checked in and those should always be used to build the project.
Because then the exact defined version for which the Build was designed for and is tested with and known working with is used forever and the build will always behave the same, no matter what changes in later Gradle versions.
If you of course install an arbitrary Gradle version and use that to run a build that was not designed for that version, it of course may break.

> It is a big step away from standard declarative builds toward ad hoc imperative ones

Well, you can write imperative builds with Gradle, you have that flexibility. But you shouldn't and it is not idiomatic. An idiomatic Gradle build is at least as declarative as a Maven build if it is properly written. Any imperative logic should be "hidden" in custom plugins or tasks which are super easy to create with Gradle opposed to with Maven.

It's basically what Uncle Ben always says: "With great power comes great responsibility".

ctrueden:
Thanks @elect86, that makes sense about the Gradle wrapper, and preferring declarative over imperative. I think I'd like Gradle better if I were the one making the Gradle files. 😜 Right now my brain is comparing apples to oranges: my carefully organized `pom.xml` files which have been iterated on for years, obsessively consistent in style and formatting... versus random Gradle files I see around the web that other people probably wrote in a matter of minutes or hours, just to the point of "it works" and falling short of correctness from a best practices standpoint.

ctrueden:
@elect86 I do think Maven's use of XML with a defined schema is a big advantage over Gradle, though: I've written lots of tooling both inside and outside the JVM that consumes `pom.xml` files and can reason about them in many useful ways. As soon as you go back to "my build script is a Turing-complete program" or what-have-you, you lose this power unless you constrain the build file to a strict set of assumptions... like a schema. Does Gradle have such a thing? A constrained version of the Gradle build file that can be validated with a schema?

ctrueden:
See #72 for the revised version of this work.

PR #46, opened 2021-11-15
Data Provenance Revisions

gselzer:
Revisions requested by @ctrueden for #35


ctrueden:
I cleaned up the history so it's based on the latest main branch, squashed the failing commits, and removed WIP labels. Review will need to wait till next week, though.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -770,22 +771,19 @@ private void initOpDirectory() {
...
-	 * <ul>
-	 * <li> some {@link Hints} given to this environment via {@link #setHints(Hints)}
-	 * <li> a {@link DefaultHints}
-	 * </li>
-	 * TODO: Consider concurrency issues
-	 * @return a {@link Hints}
-	 */
-	private Hints getHints() {
+	@Override
+	public Hints getDefaultHints() {
```
ctrueden, line 842: I know this is pretty much what we had agreed to do! But I feel the naming is now confusing: we have a `DefaultHints` *class* and `getDefaultHints`/`setDefaultHints` *methods*, and the methods do not operate on `DefaultHints` objects (necessarily), but rather the interface. It's a name clash conceptually between our usage of `Default` as a prefix for the default implementation for interfaces, and "default" as a technical term meaning "the settings which, all other things being equal and unspecified, we use to operate."

It's been a long time since the hints API has taken up any of my brain space, so I may have reverted back to naivety here, but in a perfect world I think `Hints` should be a single class with no interface at all. And anyone who wants to modify how it works (which should be vanishingly rare), would just subclass it.

I'm not necessarily requesting any changes right now, just writing down my thoughts. Perhaps all of this will be moot later as the codebase continues to evolve in this and subsequent PRs.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
```diff
@@ -39,7 +39,7 @@
 	 * @return a new {@link Hints} Object with the same hints as this
 	 *         {@link Hints}
 	 */
-	public Hints getCopy(boolean generateID);
+	public Hints copy(boolean generateID);
```
ctrueden: :+1: And it's more than just *looking* nicer... it is more *correct* because `getFoo` implies you are just giving back some member or detail about the object, not _creating something new_. In other words: `thing.getFoo() == thing.getFoo()` should return `true`—or at minimum `thing.getFoo().equals(thing.getFoo())` in the cases where you need to manufacture a new object.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/hint/AbstractHints.java
```diff
@@ -78,10 +74,17 @@ protected String getPrefix(String hint) {
...
+	 * different {@code UUID}s. This should not prevent, for matching purposes,
+	 * two {@link Hints} from being equal.
+	 */
 	@Override
 	public boolean equals(Object that) {
-		if(!(that instanceof AbstractHints)) return false;
-		AbstractHints thoseHints = (AbstractHints) that;
+		if(!(that instanceof Hints)) return false;
+		Hints thoseHints = (Hints) that;
 		return getHints().equals(thoseHints.getHints());
```
ctrueden: Ahh, `hints.getHints()` is too confusing... can we please rename the `getHints()` method to `map()`? I think `theseHints.map().equals(thoseHints.map())` reads much more nicely! And it communicates what is actually being given back: a `Map` of key-value pairs.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/hint/AdaptationHints.java
```diff
@@ -28,10 +28,10 @@ public static AdaptationHints generateHints(Hints hints, boolean generateID) {
 				.getValue()));
 
 		// add Adaptation.NO
-		UUID id = generateID ? UUID.randomUUID() : hints.executionChainID();
-		AdaptationHints newHints = new AdaptationHints(id, map);
-
-		return newHints;
+		if (generateID) {
+			return new AdaptationHints(null, map);
```
ctrueden: For short contingent + single-line consequent, you can write:
```java
if (generateID) return new AdaptationHints(null, map);
```
and it is shorter. This is a matter of taste—some people hate this formulation. But I fall into the camp of liking it more, and finding it easier to read, than bracing out the single statement.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/BaseOpHints.java
```diff
@@ -12,14 +12,12 @@
 
 	public static class Simplification {
 		public static final String PREFIX = "simplification";
-		public static final String ALLOWED = PREFIX + ".ALLOWED";
```
ctrueden: :tada:

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpService.java
```diff
@@ -57,4 +58,7 @@
 	/** Retrieves the motherlode of available ops. */
 	public OpEnvironment env();
 
+	/** Retrieves a log of all Ops executed */
+	public OpHistory history();
```
ctrueden, line 62: ❤️

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/simplify/SimplifiedOpRef.java
```diff
@@ -108,6 +108,16 @@ public static SimplifiedOpRef simplificationOf(OpEnvironment env, OpRef ref,
 	 * <p>
 	 * Adaptation is similarly forbidden, as to convert most Op types to
 	 * {@link Arity1} you would need an identical copy Op.
+	 * <p>
+	 * NB We create the {@link Hints} using the construction
+	 * {@code new DefaultHints(hints.getHints())} instead of using the
+	 * construction {@code hints.copy(true)} because we <b>do not</b> want to
+	 * ensure that we are not passing a {@code SimplificationHints} to the
```
ctrueden, line 115: "we *do not* want to ensure that we are not passing a `SimplificationHints`" &rarr; "we want to ensure that we are not passing any specialized hint subclass (e.g. `SimplificationHints`)"

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/simplify/SimplifiedOpRef.java
```diff
@@ -108,6 +108,16 @@ public static SimplifiedOpRef simplificationOf(OpEnvironment env, OpRef ref,
...
 	 * {@link Arity1} you would need an identical copy Op.
+	 * <p>
+	 * NB We create the {@link Hints} using the construction
+	 * {@code new DefaultHints(hints.getHints())} instead of using the
+	 * construction {@code hints.copy(true)} because we <b>do not</b> want to
+	 * ensure that we are not passing a {@code SimplificationHints} to the
+	 * matcher. We need to forbid simplification to prevent an infinite loop, but
+	 * the Simplification hint type is restricted to
+	 * {@code Simplification.IN_PROGRESS} by the {@code SimplificationHints}. This
+	 * is likely a bug, but would be best fixed by sliming down the {@link Hints}
```
ctrueden, line 119: "sliming down" &rarr; "slimming down" 💩

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpAdaptationInfo.java
```diff
@@ -133,4 +137,37 @@ public AnnotatedElement getAnnotationBearer() {
 		return srcInfo.getAnnotationBearer();
 	}
 
+	/**
+	 * Returns the version of the adapted Op.
+	 * <p>
+	 * Note that {@code adaptorInfo.version()} is used as the Op returned is an
+	 * inner class of the adaptor Op, and will thus have the same version as the
```
ctrueden, line 145: Yep, that seems correct to me. :+1: Glad it is noted explicitly here.

scijava/scijava-ops-api/src/main/java/module-info.java
```diff
@@ -1,6 +1,7 @@
 module org.scijava.ops.api {
 
 	exports org.scijava.ops.api;
+	exports org.scijava.ops.api.features;
```
ctrueden, line 4: > it is regrettable that

Let's revisit this after the main branch has caught up to all the latest work. At that time, we'll be doing a full module+package review anyway, and this design decision will come up again.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -327,18 +320,30 @@ private Object wrapViaCache(MatchingConditions conditions,
...
 		OpCandidate candidate = findOpCandidate(conditions.ref(), conditions.hints());
 
+		instantiateAndCache(conditions, candidate);
+
+		return conditions;
+	}
+
+	private void instantiateAndCache(MatchingConditions conditions,
+		OpCandidate candidate)
+	{
```
ctrueden, line 396: The private `cacheOp` method has only one line of code, which is very simple. I would inline it here, the only place it is called.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -284,36 +282,31 @@ public void register(final OpInfo info) {
...
 
-		// cache raw Op
-		OpInstance instance = OpInstance.of(op, info, typeVarAssigns);
-		cacheOp(conditions, instance);
+		instantiateAndCache(conditions, candidate);
+
+		return conditions;
 	}
 	
 	private Object wrapViaCache(MatchingConditions conditions,
```
ctrueden: I know the `wrapViaCache` method gets simpler in later commits. But either way, the placement feels weird here sandwiched between `insertCacheHit` and `generateCacheHit`, two very related but distinct methods complex enough to require some javadoc explaining them even though they are private.

So I suggest moving `wrapViaCache` elsewhere in the class, if it's not already moot from later work.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -284,36 +282,31 @@ public void register(final OpInfo info) {
 	 * This Op instance is put into the {@code opCache}, and is retrievable via
```
ctrueden, line 331: A small thing, but instead of `{@code opCache}` I would write either:
* `{@link #opCache}` – so that the IDE knows what you are talking about, and helps you deal with any future renames better; or
* `op cache` – i.e. just call it by name with plain English, since that's what this method does.

ctrueden: And another small thing: I thought the convention when talking about ops was to call it either "an op" (small case) or `an {@link Op}` (i.e. the interface). But since many ops do not actually implement `Op`, I'd prefer the former wherever possible. OR: we can agree that we are going to capitalize Op everywhere when talking about these things, and be consistent about that. I don't feel super strongly, other than my obsession with consistency.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/ManualOpCandidate.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.scijava.ops.api.OpCandidate;
+import org.scijava.ops.api.OpEnvironment;
+import org.scijava.ops.api.OpInfo;
+import org.scijava.ops.api.OpRef;
+import org.scijava.ops.engine.matcher.OpMatcher;
+
+public class ManualOpCandidate extends OpCandidate {
```
ctrueden, line 15: It looks like the only thing this class does is provide a convenience generator for the type var assigns, no? And it does that using the provided `OpMatcher`?

Wouldn't it make more sense to have a method `default OpCandidate candidate(OpRef, OpInfo)` of `OpMatcher` that uses this logic?

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
```diff
@@ -6,16 +6,22 @@
...
 
 /**
- * A basic interface for storing and accessing Hints. The general structure for a Hint is
+ * A basic interface for storing and accessing Hints. The general structure for
+ * a Hint is
  * <p>
  * {@code hint = hintType.option}
  * <p>
  * <ul>
- * <li>{@code hintType} designates the category of hint (for example, {@code Simplification} in {@link Simplification})
```
ctrueden, line 14: `<li>` requires a closing `</li>` after each blurb.

I know, javadoc doesn't... but it's technically incorrect... the best kind of incorrect...

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
```diff
@@ -6,16 +6,22 @@
...
- * <li>{@code hint} is the combination of {@code hintType} and {@code option} with a delimiting {@code .}
+ * <li>{@code hintType} designates the category of hint
+ * <li>{@code option} designates the preference within the category
+ * <li>{@code hint} is the combination of {@code hintType} and {@code option}
+ * with a delimiting {@code .}
  * </ul>
- *
+ * <p>
+ * For example, you might want a {@code hintType} to designate preferences on
+ * lossiness (as a tradeoff for performance). That {@code hintType} might be
```
ctrueden, line 20: Suggest rewording: "For example, as a tradeoff for performance, you might want a {@code hintType} to designate preferences on lossiness." Parenthetical notes (which I also love!) are a bad habit, easily overused. 😉

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
```diff
@@ -6,16 +6,22 @@
...
+ * {@code Lossiness}, with {@option}s {@code LOSSLESS} and {@LOSSY}.
```
ctrueden, line 21: `{@option}` &rarr; `option`

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/hint/SimplificationHints.java
```diff
@@ -38,8 +38,13 @@ public String setHint(String hint) {
...
 
 	@Override
-	public Hints copy(boolean generateID) {
-		return SimplificationHints.generateHints(this, generateID);
+	public Hints copy() {
+		return SimplificationHints.generateHints(this, false);
+	}
+
+	@Override
+	public Hints copyRandomUUID() {
```
ctrueden: I was going to comment on this name, but I know that we get rid of the UUIDs for hints later, so... 👍

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/LambdaTypeBaker.java
```diff
@@ -20,7 +20,12 @@
 import javassist.CtNewMethod;
 import javassist.NotFoundException;
 
-public class LambdaTypeBaker {
+public final class LambdaTypeBaker {
+
+	/**
+	 * Prevents instantation of this public utility class
```
ctrueden, line 26: `instantation` &rarr; `instantiation`

And as an aside, I'd like to standardize how we are executing the utility class design pattern across the codebase.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpMethodInfo.java
```diff
@@ -328,14 +328,17 @@ private String createFunctionalMethod(Method m) {
...
+	 */
 	private String getParameterName(Class<?> c) {
-		if (!c.isArray()) return c.getSimpleName();
-		// TODO: if c is an array, simpleName will include brackets (which is
-		// illegal in a class name). To differentiate Object[] from Object, we map
-		// Object[] to ObjectArr. This is not truly extensible, since someone
-		// could create an ObjectArr class which might conflict, so it would be
-		// best to find a better solution.
-		return  c.getComponentType().getSimpleName() + "Arr";
+		return c.getSimpleName().replaceAll("[^a-zA-Z0-9_]", "_");
```
ctrueden, line 349: Here's a more general way to do this:
```java
public static String toIdentifier(String s) {
	StringBuilder sb = new StringBuilder();
	s.chars().forEach(codePoint -> {
		if ((sb.length() == 0 && Character.isJavaIdentifierStart(codePoint)) ||
			(sb.length() > 0 && Character.isJavaIdentifierPart(codePoint)))
		{
			sb.append("_"); // Use underscore for invalid character.
		}
		else sb.appendCodePoint(codePoint);
	});
	return sb.length() == 0 ? "_" : sb.toString();
}
```
This assumes that the goal, as stated in the javadoc, is really to form a valid Java identifier from a string. Valid Java identifier are allowed to have many more characters than only `A-Za-z0-9_`. They also cannot start with `0-9`, hence the `isJavaIdentifierStart` versus `isJavaIdentifierPart` distinction. But Java class names are always valid identifiers anyway; it's just the array thing that throws us for a loop here. So the above is probably overkill.

ctrueden: I also thought about whether we could replace square bracket symbols with other Unicode approximations, but tragically, I couldn't find any that are valid Java identifiers.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpClassInfo.java
```diff
@@ -65,17 +65,19 @@
...
 	private final Class<?> opClass;
+	private final String version;
 	private Struct struct;
 	private ValidityException validityException;
 	private final double priority;
 	private final Hints hints;
 
-	public OpClassInfo(final Class<?> opClass, final String... names) {
-		this(opClass, priorityFromAnnotation(opClass), names);
+	public OpClassInfo(final Class<?> opClass, final String version, final String... names) {
```
ctrueden, line 73: Why not keep the old constructor as well, and pass `VersionUtils.getVersion(opClass)` as the version to a `this(...)` call?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpFieldInfo.java
```diff
@@ -61,15 +61,17 @@
...
+	private final String version;
 	private final List<String> names;
 
 	private Struct struct;
 	private ValidityException validityException;
 
 	private final Hints hints;
 
-	public OpFieldInfo(final Object instance, final Field field, final String... names) {
+	public OpFieldInfo(final Object instance, final Field field, final String version, final String... names) {
```
ctrueden, line 71: Same: the simpler constructor can have the default version calculation behavior.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpMetadata.java
```diff
@@ -0,0 +1,43 @@
+package org.scijava.ops.api;
+
+import java.lang.reflect.Type;
+import java.util.UUID;
+
+public class OpMetadata {
```
ctrueden, line 5: I love this class. It's important, and needs javadoc.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpMetadata.java
```diff
@@ -0,0 +1,43 @@
...
+
+	public Hints hints() {
+		return hints;
+	}
+
+	public UUID executionID() {
+		return executionID;
+	}
+
+	public Type type() {
```
ctrueden, line 33: The fields and accessor methods ought to be ordered the same.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpWrapper.java
```diff
@@ -2,13 +2,12 @@
...
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.UUID;
 
 import org.scijava.plugin.SciJavaPlugin;
 import org.scijava.types.Types;
 
 public interface OpWrapper<T> extends SciJavaPlugin {
-	T wrap(T op, OpInfo info, Hints hints, OpHistory history, UUID uuid, Type reifiedType);
+	T wrap(T op, OpMetadata metadata);
```
ctrueden: So much cleaner! 👍

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/RichOp.java
```diff
@@ -0,0 +1,21 @@
+package org.scijava.ops.api;
+
+import java.util.UUID;
+
+import org.scijava.types.GenericTyped;
+
+public interface RichOp extends GenericTyped {
```
ctrueden: Ahh, this interface has the same API as `OpMetadata`. That feels bad to me, since neither extends the other. Should `OpMetadata` implement `RichOp`? That wouldn't make sense, because `OpMetadata` doesn't have the pre/post processing methods. Should `OpMetadata` be an interface, which `RichOp` extends? That seems closer to correct to me. But then the current `OpMetadata` class would become `DefaultOpMetadata`, right? And `DefaultRichOp` could extend it?

I know, composition over inheritance and all that... it does make sense that `RichOp` _has_ an `OpMetadata`... but still, they have all the same methods! At that point, there should be an interface they both extend. OR: `RichOp` could have a single `metadata()` method that returns the `OpMetadata`, and then it wouldn't have to recapitulate all the same methods directly—you'd just write e.g. `richOp.metadata().info()`. But I don't like the verbosity.

@gselzer What do you think?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/DefaultRichOp.java
```diff
@@ -1,46 +1,29 @@
...
-import org.scijava.ops.api.Hints;
 import org.scijava.ops.api.OpExecutionSummary;
-import org.scijava.ops.api.OpHistory;
-import org.scijava.ops.api.OpInfo;
 import org.scijava.ops.api.OpMetadata;
 import org.scijava.ops.api.RichOp;
 import org.scijava.ops.api.features.BaseOpHints.DependencyMatching;
 
-public class DefaultRichOp implements RichOp {
+public abstract class DefaultRichOp implements RichOp {
```
ctrueden: Is there a reason `RichOp` needs to be an interface? Nothing implements it except for this abstract class, right? I'd favor just calling it `public abstract class RichOp` directly and being done with it. Do people really need to be able to extend this? (Even if so: they can still extend—just not compose.)

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/RichOp.java
```diff
@@ -1,21 +1,23 @@
...
 
-	OpHistory history();
-
-	OpInfo info();
+	Object op();
 
-	UUID executionID();
-
-	Hints hints();
+	OpMetadata metadata();
```
ctrueden, line 33: OK, I see you changed it here along the lines of what I wrote previously. So ignore my previous comments about `OpMetadata` maybe becoming an interface.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -67,9 +67,9 @@ public void testPriorityProvenance() {
 		List<OpExecutionSummary> history1 = ops.history().executionsUpon(out1);
 		List<OpExecutionSummary> history2 = ops.history().executionsUpon(out2);
 		Assert.assertEquals(1, history1.size());
-		Assert.assertEquals(baz, history1.get(0).executor());
+		Assert.assertEquals(baz, history1.get(0).op().op());
 		Assert.assertEquals(1, history2.size());
-		Assert.assertEquals(bar, history2.get(0).executor());
+		Assert.assertEquals(bar, history2.get(0).op().op());
```
ctrueden: `.op().op()`? What? I have ideas here. 😆 

So, the first `.op()` gives the `RichOp`, while the second `.op()` gives the backing instance. Here's my radical proposal:

* We don't need the `Op` marker interface anymore, right? Because we ditched SJC2 and `@Plugin`? So then we aren't using it?
* That frees up the name, so we could... drum roll... rename `RichOp` to `Op`!
* And change the `op()` method of `RichOp` (now `Op`) to `instance()`.
* So then the above would read: `history2.get(0).op().instance()`

What do you think? Would it be feasible?

If we can swing it, I think it'd be a really big step forward, naming-wise.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpWrapper.java
```diff
@@ -7,7 +7,7 @@
 import org.scijava.types.Types;
 
 public interface OpWrapper<T> extends SciJavaPlugin {
-	T wrap(T op, OpMetadata metadata);
+	RichOp<T> wrap(T op, OpMetadata metadata);
```
ctrueden: Of course, if we rename `RichOp` to `Op`, we'll have to change all the places we refer to the various instances as `op`. But whether we rename `RichOp` or not, I feel we need to be more consistent and clear: either `op` means the _instance_, or `op` means the _`RichOp` wrapper_. It should not mean both! So if we don't rename `RichOp` to `Op`, then the `.op()` accessor in the history should _not_ return a `RichOp`. I guess the reason we are getting away with this is because we bend over backwards to make sure that the `RichOp` instance is also an instance of the same functional interface as the wrapped op, right?

And then there's `OpInstance`... yet another thing. I know we can do better with these names... somehow. Let's discuss when we have time.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpHistory.java
```diff
@@ -10,10 +10,9 @@
  * Log describing each execution of an Op. This class is designed to answer two
  * questions:
  * <ol>
- * <li>Given an {@link Object} output (e.g. a {@code List<String>}), what Op(s)
- * mutated that output?
- * <li>Given an {@link Object} op, what {@link OpInfo}s were utilized to
- * implement that Op's functionality?
+ * <li>What Op(s) produced and/or mutated this output?
```
ctrueden, line 10: Missing `</li>`

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
```diff
@@ -26,15 +27,23 @@
...
 
-	boolean containsHint(String hint);
+	boolean contains(String hint);
 
-	boolean containsHintType(String hintType);
+	default boolean containsNone(String... hints) {
+		return !containsAny(hints);
+	}
 
-	Map<String, String> getHints();
```
ctrueden: Oh, `getHints` went away. So ignore my comments about renaming it to `map()`, then.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/hints/SimplificationHintTest.java
```diff
@@ -55,7 +55,7 @@ public void testSimplificationPerCallHints() {
 			"test.simplification.hints").inType(Integer[].class).outType(
 				Integer[].class).function(hints);
 		// make sure we cannot find the Op when adaptation is not allowed
-		hints.setHint(Simplification.FORBIDDEN);
+		hints = hints.plus(Simplification.FORBIDDEN);
```
ctrueden, line 56: Ah, if only Java had operator overloading. 😄

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
```diff
@@ -0,0 +1,43 @@
+
+package org.scijava.ops.api;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class InfoChain {
```
ctrueden, line 15: Chain implies linear. This is actually a tree. So I'd name it `OpTree`.

Alternately: this tree is a hierarchy of *resolved* dependencies, right? So it could be named `ResolvedOp` or something like that. We can discuss more later. I know I'm talking about naming a lot, but this stuff is so complex, that good names are the only way to avoid people becoming totally overwhelmed by the abstractions.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
```diff
@@ -6,10 +6,20 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
+/**
+ * A data structure wrangling a hierarchy of {@link OpInfo}s.
```
ctrueden, line 11: More javadoc, please. When are these created? Every time you execute an op? Every time you match an op?

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
```diff
@@ -21,23 +31,55 @@ public InfoChain(OpInfo info, List<InfoChain> dependencies) {
...
-		throw new UnsupportedOperationException("Not yet implemented");
+	/**
+	 * Builds a String uniquely identifying this tree of Ops. As each
+	 * {@link OpInfo#id()} should be unique to <b>that</b> {@link OpInfo}, we can
+	 * use those to uniquely identify a hierarchy of {@link OpInfo}s, and thus by
+	 * extension a unique {@link InfoChain}.
+	 *
+	 * @return a {@link String} uniquely identifying this {@link InfoChain}
+	 */
+	public String signature() {
```
ctrueden, line 57: :+1:

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpMetadata.java
```diff
@@ -6,12 +6,12 @@
 public class OpMetadata {
 
 	private final Type type;
-	private final OpInfo info;
+	private final InfoChain info;
```
ctrueden, line 8: Maybe `MatchedOp` would be a better name? Or `ResolvedOp`. Something that conveys that this is an `OpInfo`, but with resolved dependencies.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpInstance.java
```diff
@@ -13,31 +13,31 @@
  * 
  * @author Gabriel Selzer
  */
-public class OpInstance implements GenericTyped {
+public class OpInstance<T> implements GenericTyped {
```
ctrueden: I'd like javadoc on the relationship between `OpInfo`, `InfoChain` (or whatever it ends up being called), `OpInstance`, and `RichOp`. The terms 1-to-1 and 1-to-many are very useful to quickly convey how these things get instantiated. Also "is a" and "has a", of course.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
```diff
@@ -65,11 +66,20 @@ public OpInfo info() {
 		return info;
 	}
 
-	public Object op() {
+	public OpInstance<?> op() {
+		return OpInstance.of(generateOp(), this, info.opType());
```
ctrueden, line 67: Assuming the relationship between `InfoChain` and `OpInstance` is 1-to-many—that is, that an `InfoChain` can generate multiple `OpInstance` objects—we may want to call this method `instantiate()` instead. Much clearer than `op()`, which might simply be an accessor. As discussed elsewhere, `.op()` can currently refer to too many different things.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpWrapper.java
```diff
@@ -7,7 +7,7 @@
 import org.scijava.types.Types;
 
 public interface OpWrapper<T> extends SciJavaPlugin {
-	RichOp<T> wrap(T op, OpMetadata metadata);
+	RichOp<T> wrap(OpInstance<T> op, OpMetadata metadata);
```
ctrueden, line 10: At this point, I'm going to punt on the "op" terminology (`OpInfo`, `OpInstance`, `RichOp`, and `InfoChain`) until after the entire PR chain is merged. Then we can assess and finalize the naming. I filed scijava/scijava#81 for this.

scijava/scijava-ops-engine/templates/main/java/org/scijava/ops/engine/matcher/impl/OpWrappers.vm
```diff
@@ -26,25 +27,25 @@ public class OpWrappers {
 	public static class ProducerOpWrapper<T> implements OpWrapper<Producer<T>> {
 
 		@Override
-		public RichOp<Producer<T>> wrap(final Producer<T> op, //
+		public RichOp<Producer<T>> wrap(final OpInstance<Producer<T>> instance, //
```
ctrueden, line 30: I'm tempted to rename this function to `enrich` since its job is to go from `OpInstance` to `RichOp`. 😉

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpHistory.java
```diff
@@ -42,7 +39,7 @@
 	// -- DATA STRCUTURES -- //
 
 	/**
-	 * {@link Map} responsible for recording the {@link Graph} of {@link OpInfo}s
+	 * {@link Map} responsible for recording the {@link InfoChain} of {@link OpInfo}s
```
ctrueden, line 42: I am glad we ditched the `Graph` dependency. :+1:

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultInfoChainGenerator.java
```diff
@@ -66,7 +66,7 @@ public boolean canGenerate(String signature) {
 
 	@Override
 	public double priority() {
-		return Double.MIN_VALUE;
+		return -(1e10);
```
ctrueden: Haha, good ol' useless `Double.MIN_VALUE`. What people usually mean when they write that is `-Double.MAX_VALUE`. But I'm glad we're not using that here, because it's always possible someone has a valid reason to want their thing at an even lower priority, so it's good to leave room.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -198,4 +226,23 @@ private OpInfo singularInfoOfName(String name) {
...
+	public final Computers.Arity1<Double[], Double[]> op = (in, out) -> {
+		for (int i = 0; i < in.length && i < out.length; i++)
+			out[i] = in[i];
+	};
+
+	@Test
+	public void testAdaptatorWithDependencies() {
+		Function<Double[], Double[]> f = ops.op("test.provenanceComputer").inType(Double[].class).outType(Double[].class).function();
+		@SuppressWarnings("unchecked")
+		InfoChain chain = ((RichOp<Function<Double[], Double[]>>) f).infoChain();
```
ctrueden, line 294: Maybe `OpBuilder` should have convenience methods like `richFunction()` and `richComputer()`? Or: it could return `OP` where `<OP extends T & RichOp<T>>`, so the caller can assign to either. Not sure how much pain that would solve versus introduce, though.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChainGenerator.java
```diff
@@ -38,23 +38,27 @@ InfoChain generate(String signature, Map<String, OpInfo> idMap,
...
 			.filter(g -> g.canGenerate(signature)) //
-			.max((i1, i2) -> (int) (i1.priority() - i2.priority()));
-		return gen;
+			.collect(Collectors.toList());
+		if (suitableGenerators.size() == 0) throw new IllegalArgumentException(
+			"No InfoChainGenerator in given collection " + generators +
+				" is able to reify signature " + signature);
+		if (suitableGenerators.size() > 1) throw new IllegalArgumentException(
+			"Signature " + signature +
+				" is able to be reified by multiple generators: " + suitableGenerators);
```
ctrueden, line 52: If I'm reading this right: you got rid of generator priority, and now assume that only one of the discovered generators will handle each case. If someone writes a plugin that has any overlap with existing generation cases, it will break the system because we'll hit this case of more than one suitable generator.

In this case, I would consider dropping the extensibility. There is an argument to be made to keep it, if new sorts of generation need to be accommodated, and people take care to only handle their newly added cases unambiguously. But if we don't have a clear vision for any new generation scenarios, we may want to simply hardcode the list of generators to the ones we provide, and leave it at that. Could make this code simpler.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
```diff
@@ -21,6 +21,8 @@
  */
 public interface OpInfo extends Comparable<OpInfo> {
 
+	static final String IMPL_DECLARATION = "|Vanilla:";
```
ctrueden, line 24: Vanilla? I love this slang and use it all the time, but we may want to use something more easily understood by non-English native speakers.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -195,10 +195,8 @@ public void testOpWithDependencyRecoveryFromString() {
 	@Test
 	public void testAdaptationRecoveryFromString() {
 		Function<Double[], Thing[]> f = ops.op("test.provenanceMapped").inType(Double[].class).outType(Thing[].class).function();
+		String signature = ops.history().signatureOf(f);
```
ctrueden, line 225: ❤️

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -260,15 +288,16 @@ private OpInfo singularInfoOfName(String name) {
...
 		String signature = chain.signature();
 		Nil<Function<Double[], Double[]>> special = new Nil<>() {};
-		Nil<Double[]> inType = Nil.of(Double[].class);
-		Nil<Double[]> outType = Nil.of(Double[].class);
-		Function<Double[], Double[]> actual = ops.env().opFromID(signature, special, new Nil[] {inType}, outType);
-		Double[] apply = actual.apply(new Double[] {1., 2., 3.});
+		Function<Double[], Double[]> actual = ops.env().opFromSignature(signature,
+			special);
+		@SuppressWarnings("unused")
+		Double[] apply = actual.apply(new Double[] { 1., 2., 3. });
```
ctrueden, line 300: Dontcha wanna assert the result? Or really... instead of calling `apply`, assert that the `actual` object is (or wraps) the expected function?

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -260,15 +288,16 @@ private OpInfo singularInfoOfName(String name) {
 
 	@Test
 	public void testAdaptatorWithDependencies() {
```
ctrueden, line 290: `Adaptator` &rarr; `Adaptation`? Or `Adapter`?

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/hints/AdaptationHintTest.java
```diff
@@ -81,8 +79,7 @@ public void testNonAdaptableOp() {
 			ops.op("test.adaptation.unadaptable").inType(Double[].class).outType(
 				Double[].class).computer();
 			throw new IllegalStateException("This op call should not match!");
-		} catch( IllegalArgumentException e) {
-			assertTrue(e.getCause() instanceof OpMatchingException);
+		} catch( OpMatchingException e) {
```
ctrueden, line 82: Instead of an empty catch block, there should be a comment here saying something like:
```java
// NB: Expected behavior.
```

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/features/OpMatchingException.java
```diff
@@ -1,4 +1,4 @@
-package org.scijava.ops.engine.matcher;
+package org.scijava.ops.api.features;
 
 import org.scijava.ops.api.OpRef;
```
ctrueden, line 4: I know that `OpMatchingException` extends `RuntimeException` for downstream convenience—being able to pretend that everything is always going to be OK, and not forcing people to always catch the `OpMatchingException`—but I wonder if this is the wrong choice. From the Oracle Java documentation: "If a client can reasonably be expected to recover from an exception, make it a checked exception. If a client cannot do anything to recover from the exception, make it an unchecked exception." Is it reasonable to "recover from" a failed matching attempt? I would say yes, in the same way it's reasonable to recover from e.g. a non-existent file on the file system. Perhaps we could hedge our bets a bit, and make the matching API in `OpEnvironment` throw `OpMatchingException`, which would become checked, while `OpBuilder` could do something different, such as returning an `Optional` or null.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/simplify/SimplifiedOpInfo.java
```diff
@@ -183,10 +183,9 @@ private static double calculatePriority(OpInfo srcInfo, SimplificationMetadata m
 			LossReporter<T, R> op = env.op("lossReporter", specialTypeNil, new Nil[] {
 				Nil.of(nilFromType), Nil.of(nilToType) }, Nil.of(Double.class));
 			return op.apply(from, to);
-		} catch(IllegalArgumentException e) {
-			if (e.getCause() instanceof OpMatchingException)
```
ctrueden, line 177: Definitely better!

ctrueden (approved):
Whew! That one took a while. There were a lot of intermediate commits with subsequent changes later down the chain. I also reviewed commits out of order because GitHub defaults to date order rather than commit order in the commit view. 😾

I filed an issue for the most major issue: the naming of "op"/`OpInfo`/`OpInstance`/`InfoChain`/`RichOp`. For the rest, I'll do one more quick scan and file issues as appropriate any followup action items.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
ctrueden: scijava/scijava#82

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/hint/AbstractHints.java
ctrueden: I believe this was already addressed in a later commit.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
ctrueden: scijava/scijava#83

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/hints/AdaptationHintTest.java
ctrueden: scijava/scijava#84

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/features/OpMatchingException.java
ctrueden: scijava/scijava#85

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/RichOp.java
ctrueden: I believe this issue is moot, because `RichOp` now has a `metadata()` accessor that obtains the associated `OpMetadata`.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
ctrueden: scijava/scijava#81

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpWrapper.java
ctrueden: scijava/scijava#81

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
ctrueden: scijava/scijava#81

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpMetadata.java
ctrueden: https://github.com/scijava/scijava/issues/81#issuecomment-986998461

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpInstance.java
ctrueden: https://github.com/scijava/scijava/issues/81#issuecomment-987171791

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
ctrueden: https://github.com/scijava/scijava/issues/81#issuecomment-987171791

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
ctrueden: scijava/scijava#86

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChainGenerator.java
ctrueden: scijava/scijava#87

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
ctrueden: scijava/scijava#81

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpHistory.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/DefaultRichOp.java
ctrueden: https://github.com/scijava/scijava/issues/81#issuecomment-987189209

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpMetadata.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpFieldInfo.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpClassInfo.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/LambdaTypeBaker.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/ManualOpCandidate.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/simplify/SimplifiedOpRef.java
ctrueden: scijava/scijava#88

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/simplify/SimplifiedOpRef.java
ctrueden: scijava/scijava#88

ctrueden:
OK, everything significant from the review has been filed as issues on the project board. Merge time! 🏎️💨

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: I'll just solve this one as its own issue, it won't be part of the `scijava/scijava-ops-engine/data-provenance-revisions-revisions` branch.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/hint/AdaptationHints.java
gselzer: This whole class is gone, so I'm going to mark this as resolved.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpService.java
gselzer: Too bad `OpService` will be gone soon 😆

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/simplify/SimplifiedOpRef.java
gselzer: Fixed via ce84afc8

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/simplify/SimplifiedOpRef.java
gselzer: Fixed by ce84afc8

scijava/scijava-ops-api/src/main/java/module-info.java
gselzer: That seems fine. See scijava/scijava#91

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: Fixed in 1d7afaab

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: Fixed via 0d9688f0

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: The first part is fixed via fd6ec387

gselzer: The second part is fixed via f81c3fcf

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/ManualOpCandidate.java
gselzer: @ctrueden Let me know what you think of daf0d720

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
gselzer: See 343d4f61.

I'd like to get a regex that can fix this project-wide. Haven't succeeded though ☹️

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
gselzer: 343d4f61

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
gselzer: 343d4f61

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/LambdaTypeBaker.java
gselzer: See 5a0b0b14 for the typo fix.

> And as an aside, I'd like to standardize how we are executing the utility class design pattern across the codebase.

@ctrueden what would you like to standardize? The fact that we do it, I assume? Can you make an issue specifying what you'd like standardized?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpMethodInfo.java
gselzer: I'd be happy to use this, with the caveat that it goes in some general SciJava utility class. This seems much broader than Ops. We could add this to [`StringUtils`](https://github.com/scijava/scijava-common/blob/master/src/main/java/org/scijava/util/StringUtils.java)?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpClassInfo.java
gselzer: See e8fb5e24

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/OpFieldInfo.java
gselzer: See e8fb5e24

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
gselzer: Succeeded using @ctrueden's expertise, via dd5ff001 😁

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpMetadata.java
gselzer: Done with 57d3c845

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpMetadata.java
gselzer: Done with 57d3c845

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpHistory.java
gselzer: Fixed with dd5ff001

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
gselzer: > Chain implies linear. This is actually a tree. So I'd name it `OpTree`.

@ctrueden I like `OpInfoTree`, or maybe just `InfoTree`. We shouldn't name it `OpTree`, because it has no Ops; it only has `OpInfo`s. 

> this tree is a hierarchy of _resolved_ dependencies, right?

I'm not sure what you mean by this. If you mean that each `OpDependency` is represented by an `OpInfo` in the right place, then yeah, it is.

I added some javadoc to `InfoChain` in f7cd642c8de8b818498ed06bb6725da3e0faa93d.

gselzer:
I was able to resolve a good amount of these comments on the `scijava/scijava-ops-engine/data-provenance-revisions-revisions` branch! What a mouthful.

I'll continue with the revisions next week.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
gselzer: Uh, not quite. In https://github.com/scijava/incubator/commit/572a1141feff553b3822c921b264b69629585b5a we make `Op` an interface that SciJava Ops Engine uses to discover Op implementations.

Note that I realized that `Op` still extends `SciJavaPlugin`, so I remove SciJava Ops API's dependency on SJC in e21e81e7

Let's just resolve these concerns in scijava/scijava#81

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
gselzer: See 0ceaa11c

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/DefaultRichOp.java
gselzer: Doing so might cause a bit of a packaging nightmare.

It is the `OpWrapper`'s job to return a `RichOp` from an `OpInstance`. `OpWrapper` needs to live in `org.scijava.ops.api`, thus we need `RichOp` in SciJava Ops API.

If we made this change, we would need to move `DefaultRichOp` (or `AbstractRichOp`, as it is now known) to SciJava Ops API. This is no good, because it will bring in additional dependencies. (for example, looking at `AbstractOpEnvironment`, we'd impose a dependency on SciJava Progress.)

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpInstance.java
gselzer: Let me know what you think of 43713f2c

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChain.java
gselzer: Let me know what you think of b2a943fe

scijava/scijava-ops-engine/templates/main/java/org/scijava/ops/engine/matcher/impl/OpWrappers.vm
gselzer: Ooh! I like that!

gselzer: *But* should we stray from the connotations that "wrap" provides? While I like `enrich`, I don't like the class name `OpEnricher` as much. That sounds like we are mutating the provided Object, but we are returning a wrapper object.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpHistory.java
gselzer: 🥳

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultInfoChainGenerator.java
gselzer: 😅

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
gselzer: Resolving this in favor of that issue.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
gselzer: 😁

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
gselzer: Fixed in ad926961

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
gselzer: Resolving this in favor of that issue

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
gselzer: Fixed with ac8ea5ac

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/hints/AdaptationHintTest.java
gselzer: Fixed this one in 98850ae3. We can catch (hah) the rest of them in a dedicated response to that issue.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/features/OpMatchingException.java
gselzer: Resolving this, since we have the issue.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpWrapper.java
gselzer: See the comment above.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpMetadata.java
gselzer: See the comment above.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/InfoChainGenerator.java
gselzer: Resolving this in favor of that issue.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/LambdaTypeBaker.java
gselzer: I created scijava/scijava#127 to track @ctrueden's desire to standardize the utility class design pattern, and as such am resolving this review comment.

PR #48, opened 2021-11-15
Rewrite SciJava Ops Engine Ops as Services

gselzer:
This PR is targeted at rewriting all Ops contained within SciJava Ops Engine as Services, discoverable using the `Discoverer.using(ServiceLoader::load)` syntax. It will be the first in multiple PRs that work towards removing the SciJava Common dependency (and admittedly goes beyond its scope in that pursuit 😅 ). Major changes include:
* The refactoring of SciJava Type's `TypeService` into a `TypeReifier`. Instead of using the SciJava `PluginService` to discover `TypeExtractor`s, it uses one or more `Discoverer`s to discover them.  This allows us to remove the SciJava Common dependency from SciJava Types 🥳
* The introduction of SciJava Log as an Incubator module. Arguably it should be called SciJava Log2?? Regardless, this partitions out [`org/scijava/log`](https://github.com/scijava/scijava-common/tree/master/src/main/java/org/scijava/log) from SciJava Common, and refactors `LogService` into `LoggerFactory`. It still has the SciJava Common dependency though, which sucks.
* The "turning-on" of ImageJ Ops2's tests. We could cherry-pick the commit to another branch if we wanted...
* The replacement of the uber-`OpEnvironment` being constructed with many `Discoverers` in each test with an `OpEnvironment` that only runs with a `StaticDiscoverer` containing needed Ops. This nearly halved testing time on my machine.
* The refactoring of all Ops in SciJava Ops Engine into Services (`Op`s or `OpCollection`s) that are retrievable through `ServiceLoader`. Regression tests were added to ensure that all written Ops are discoverable from their designated `Discoverer`. 
* The `Plugin` discovery logic was moved into SciJava Discovery Plugin. We could delete it, but it really doesn't *hurt* anything to leave it...

**TODO before merge**:
* [ ] Unfortunately, the test refactoring required me to turn off the subpackage dependence and package cycle enforcer rules; for this reason, this PR should not be merged until the tests do not break those rules. scijava/scijava#80 should be solved to fix this
* [ ] Remove SciJava Common dependency from SciJava Log
* [x] Update name to SciJava Log2?
* [ ] Consider throwing `LoggerFactory` impls into an internal package? I actually don't feel strongly about this, since I could see the user wanting to select a particular implementation, but I also don't want to expose implementations...

PR #49, opened 2021-11-15
Use Therapi Discovery in ImageJ Ops2 

gselzer:
This PR follows #48 as the second PR designed to purge SciJava Common 2 from the Incubator projects. It:
* Refactors `@Plugin` definitions as `@implNotes`, consistent with the methodology described in #44.
* Removes the vast majority of, if not all, `@Plugin` and `Priority` `import`s from the ImageJ Ops2 module.


ctrueden (approved):
Looks good. Thank you for documenting the perl commands used. That could conceivably come in handy later (or at least educate someone on how to do this sort of work).

I squashed your import cleaning commits, and did my own import cleaning pass, since there were still many unused ones left.

I will note that I got some nondeterministic test failures on my system, where ops or their dependencies sometimes could not be matched. I am inclined to blame Eclipse, even though I tried to have the projects closed in Eclipse any time I ran tests from the command line. Anyway, the last couple of rounds of testing built successfully for me, so I'm going to move ahead with this.

PR #50, opened 2021-11-16
Convert SciJava Types Tests to JUnit 5

gselzer:
This PR does what it says above.

The POM only listed `org.junit` and `org.junit.jupiter.api` as dependencies, which is very strange. Furthermore, it would only run JUnit 4 tests, which represented very few of this project's tests. This PR replaces `org.junit` with `org.junit.jupiter.engine`, and converts all tests still running JUnit 4.

TODO:
* [ ] Merge #49, the PR this one is built on top of.

### DIFF
--- a/scijava/scijava-types/pom.xml
+++ b/scijava/scijava-types/pom.xml
@@ -116,13 +116,13 @@
 
 		<!-- Test scope dependencies -->
 		<dependency>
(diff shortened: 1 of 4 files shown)


ctrueden (approved):
Great! One component down, eight more to go. :sweat_smile:

PR #51, opened 2021-11-29
Improve Discovery Framework for use in Ops API

### LINKED ISSUE #80: Add OpEnvironment API to Generate OpInfos from a Collection/Array of Classes
We need to replace the functionality of SciJava Common's plugin framework. The best way I can think of is to use `ServiceLoader` to declare some `OpEnvironmentFactory` implementations, which are capable of returning an `OpEnvironment`. This `OpEnvironmentFactory` should have:
* A method `OpEnvironment create()`
* Some prioritization mechanism allowing us to choose a single `OpEnvironment` when given multiple `OpEnvironmentFactory` objects without having to hardcode an implementation class.

The problem with this implementation is that it becomes extremely difficult to modify the `OpEnvironment`. Currently, the only functionality available is `register(OpInfo info)`, which is not helpful when we will no longer have access to the `OpInfoGenerators` and `Discoverer`s created by the `OpEnvironmentFactory`.

What we need is a more powerful API for adding Ops to the `OpEnvironment`. For now, I like the idea of 

```java
public Iterable<OpInfo> registerFrom(Class<?>... classes);
```
Such functionality should be able to parse a set of `OpInfo`s from each of the elements of `classes` and insert them into the op directory of the `OpEnvironment`. Issues to implementing such API:

* The `OpInfoGenerator`s cannot generate a `List<OpInfo>` from a `Class`, but only from a `Discoverer` (which it holds as its state). `OpInfoGenerator`s should instead have a (static?) method that takes a `Collection<? extends Class<?>>` and returns a `List<OpInfo>`. `OpInfoGenerator`s use the functionality of `Discoverer`, however, so that must be reconciled.
* `DefaultOpEnvironment` lazily initializes its `opDirectory`, which causes a nightmare when trying to add new Ops. Currently, any call to `register(OpInfo)` will populate the entire `opDirectory` (which, depending on the `Discoverer`s available, can take a while). We need to refactor the `DefaultOpEnvironment` to be smarter.
* `Discoverer` has two different methods, `implsOfType()` and `elementsTaggedWith()`. We need a wrapper that can combine the returns from both methods, returning a `List<Class<?>>` that can be passed to the `OpInfoGenerator`s.

For now, to continue development, we can use the `HACK` of https://github.com/scijava/incubator/commit/67bef7491e30b210b076f10dad20e7ae43c9fffd

gselzer:
This PR greatly simplifies the `Discoverer` API, as well as adding some new functionality via static `Discoverer` methods. The highlights are listed below:
* `Discoverer.all()`: Returns all `Discoverer`s provided as `Service`s, in addition to a `Discoverer` that is backed by `ServiceLoader`.
* `Discoverer.union(Iterable<Discoverer> itr)`: allows the combination of multiple `Discoverer`s into one uber-`Discoverer`
* `Discoverer.onlyFor(Class<?>... classes)`: allows the usage of `Discoverer` for a subset of classes, very useful for backend implementation discovery (i.e. if you want to run an `OpEnvironment` with all known `MatchingRoutine`s, but not all known `Op`s. `Discoverer.except()` provides the reverse.
These methods allow for powerful syntax like
```java
		Discoverer serviceLoading = Discoverer.using(ServiceLoader::load) //
				.onlyFor( //
						OpWrapper.class, //
						MatchingRoutine.class, //
						OpInfoGenerator.class, //
						InfoChainGenerator.class //
				);

```
This syntax is used in `AbstractTestEnvironment` to get all known `OpWrapper`s, `MatchingRoutine`s, `OpInfoGenerator`s, and `InfoChainGenerator`s, and nothing else. This allows extending tests to obtain all of these useful backend implementations while removing the burden of being slowed down by all known Ops. 

The end goal of this PR was to make it easy for an `OpEnvironment` to take an `Object` and extract `OpInfo`s from it. To this end, the method `makeDiscoverable(Object... objects)` was added to `OpEnvironment`. 

TODO:
* [ ] It seems that *almost* every incubator build, at least one *random* ImageJ Ops2 test fails due to a matching error. I haven't been able to replicate this issue with any particular test. Should this be a blocker to merging?

Closes scijava/scijava#80


scijava/scijava-discovery-therapi/src/main/java/org/scijava/discovery/therapi/TherapiDiscoverer.java
```diff
@@ -30,113 +15,73 @@
...
-		else if (file.getPath().endsWith(".jar")) {
+		} else if (file.getPath().endsWith(".jar")) {
 			try {
 				for (String s : getJarContent(path))
 					if (s.endsWith("__Javadoc.json")) {
 						filesList.add(new File(s));
 					}
-			}
-			catch (IOException exc) {
+			} catch (IOException exc) {
```
ctrueden, line 79: What's going on with the curly braces here? Are we changing code styles?

ctrueden (approved):
I am happy with this as a big step forward! I like the transition from `Discoverer` as interface required for all types of discoveries, to `Object` as primary currency.

PR #52, opened 2021-11-30
Update all incubator projects to JUnit 5

gselzer:
This PR is pretty self-explanatory. All incubator projects now run JUnit 5

Built atop the work in #50, as it is a natural continuation of that work.


ctrueden (approved):
Brilliant work, @gselzer! I just had to massage things a *liiiittle* bit to remove "WIP" from one of the commit titles, and remove a spurious remaining old JUnit import... but now it's all good!

PR #53, opened 2022-01-14
Revisions on the Revisions of the Data Provenance branch

gselzer:
Fixes to #46.


scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
```diff
@@ -20,7 +20,8 @@
  */
 public interface OpInfo extends Comparable<OpInfo> {
 
-	static final String IMPL_DECLARATION = "|Vanilla:";
+	/** Identifier for an unaltered OpInfo in an Op signature **/
+	String IMPL_DECLARATION = "|Info:";
```
ctrueden, line 24: Some reason this shouldn't be `|Op:` instead?

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Hints.java
```diff
@@ -1,53 +1,97 @@
-
```
ctrueden, line 1: A bunch of files in the repo have been Windows line endings up till this point... but now they are getting converted over one by one. This is much better done as its own commit fixing all of them at once, when this type of problem is noticed. Otherwise, the diff here is obfuscated by the line ending shift.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -337,7 +337,7 @@ private Type[] toTypes(Nil<?>... nils) {
 	/**
 	 * Creates an Op instance from an {@link OpInfo} with the provided
 	 * {@link MatchingConditions} as the guidelines for {@link OpInfo} selection.
-	 * This Op instance is put into the {@code opCache}, and is retrievable via
+	 * This Op instance is put into the {@link #opCache}, and is retrievable via
```
ctrueden, line 340: By default when you generate javadoc, private members are not included in the generated output. So then this sort of link, which goes to a private field, will not actually be functional. I am actually not sure how this ends up rendering in the default case, but we should check. If it looks bad or confusing, we can just remove the `@link` in favor writing "op cache" in normal language. But if it looks OK, I favor leaving it so that in IDEs the linkage is still there.

ctrueden (approved):
Looks good. A couple of small comments, and some commit messages do not use imperative tense as is standard, but this is good to go!

PR #47, opened 2021-11-15
View-Javadoc Revisions

gselzer:
Implements suggestions from @ctrueden in #28

PR #54, opened 2022-01-21
Revisions on the revisionson on view-javadoc

gselzer:
Addresses the revisions requested in scijava/incubator#47


scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/struct/AnnotatedOpDependencyMember.java
```diff
@@ -45,20 +45,32 @@
...
+	 * @param key the key
+	 * @param description the description
+	 * @param type the {@link Type} of this {@link Member}
+	 * @param annotation the {@link OpDependency} annotation
+	 */
 	public AnnotatedOpDependencyMember(String key, String description, Type type,
 		final OpDependency annotation)
 	{
 		this(() -> key, () -> description, type, annotation);
+		this.key = key;
```
ctrueden, line 70: This looks like more than just "cleaning" to me. A better commit message explaining what this commit actually does would have been nice.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/struct/AnnotatedOpDependencyMember.java
```diff
@@ -99,16 +113,31 @@ public boolean isAdaptable() {
...
+		description = descriptionGenerator.get();
+		descriptionGenerated = true;
 	}
 
 	@Override
 	public Type getType() {
 		return type;
 	}
 }
+
```
ctrueden, line 143: Trailing newline.

scijava/scijava-struct/src/main/java/org/scijava/struct/Member.java
```diff
@@ -16,7 +16,7 @@
 	/** Unique name of the member. */
 	String getKey();
 
-	/** Description of the member. */
+	/** Non-null description. The default description is an empty string */
```
ctrueden, line 19: Second sentence needs a period.

scijava/scijava-struct/src/main/java/org/scijava/struct/Structs.java
```diff
@@ -81,7 +81,7 @@ public static boolean checkValidity(Member<?> m, Supplier<String> name, Class<?>
 			// will be written to, but immutable parameters cannot be changed in
 			// such a manner, so it makes no sense to label them as such.
 			final String error = "Immutable " + m.getIOType() + " parameter: " +
-				name.get() + " (" + type.getName() + " is immutable)";
```
ctrueden, line 84: To be fair, we are only calling the supplier *when there is an error*. So, there is a chance that performance could be impacted. But the fact that the only supplier is just `m.getKey()` and presumably that's O(1), with no additional memory allocation... means that performance is highly unlikely to be a problem here. At least for *this* use case! But the `checkValidity` method is `public static`... should we be concerned in general? Does/will anyone else ever call this? If not: maybe we should make this private? Or if so, maybe the supplier is still a good idea?

ctrueden (approved):
A few small comments again, but merging proceeds! Keep up the good work @gselzer, you got this!

PR #56, opened 2022-01-21
Clean `ProvenanceTest`

### LINKED ISSUE #83: ProvenanceTest: assert that matched ops actually match
See https://github.com/scijava/incubator/pull/46#discussion_r763349724

gselzer:
This PR is focused on two actions, both in `ProvenanceTest`.

Firstly, it makes better assertions in the tests dealing with Op creation from a `String` signature. See https://github.com/scijava/scijava/issues/83 for the motivation behind this.

Secondly, it cleans and formats the file.

Closes scijava/scijava#83


scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -1,11 +1,7 @@
 
 package org.scijava.ops.engine.impl;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
```
ctrueden, line 4: Since when did we decide to start using wildcard imports? I don't remember discussing this.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
```diff
@@ -37,12 +38,28 @@ public interface OpInfo extends Comparable<OpInfo> {
 
 	/** Gets the op's input parameters. */
 	default List<Member<?>> inputs() {
-		return OpUtils.inputs(struct());
```
ctrueden, line 40: Post-move, does anything still use `OpUtils.inputs` to obtain the members of a struct? Or can we slim down the `OpUtils` methods henceforth? It's not in an exported anymore, right? So we could trim it?

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -246,14 +246,20 @@ public void testDependencylessOpRecoveryFromString() {
...
+				.outType(Thing.class) //
+				.function();
+		// Get the signature from the Op
+		String signature = history.signatureOf(mapper);
+		// Generate the Op from the signature and an Op type
+		Nil<Function<Double, Thing>> specialType = new Nil<>() {};
+		Function<Double, Thing> actual = ops //
+				.opFromSignature(signature, specialType);
+		// Assert Op similarity
+		Assertions.assertTrue(wrappedOpEquality(mapper, actual));
```
ctrueden, line 262: I've been wondering... we used to use `import static` with JUnit 4. Now that we are on JUnit 5, would it make sense to do the same, statically importing used `Assertions` methods? Or do you prefer the verbosity this way? It's more consistent with everything else, but I ask because it's a break from previous JUnit convention.

ctrueden (approved):
I fixed a couple of small post-rebase problems, and this should be good to go now! Just waiting for CI to give us the green checkmark first.

PR #57, opened 2022-02-11
Remove SciJava Common from Incubator Code

gselzer:
This PR removes SciJava Common as a dependency for all incubator projects, additionally removing most mentions of SciJava Common.

To allow ImageJ Ops2 access to the Ops it describes, we expose `DefaultOpEnvironment`.

TODO:
* [ ] Decide whether to clean this PR or just squash+merge. I'd prefer to squash.
* [ ] Re-enable package cycles. These were turned off to prevent failures arising from the exposure of `DefaultOpEnvironment`.

PR #58, opened 2022-03-24
Op Reduction, Rebased

gselzer:
This PR rebases the work of #32 on top of #57, making it much easier to merge.

Please see #32 for a description of the workings.


ctrueden:
@gselzer I tried to get started on this PR by:
```
gh pr checkout 58
git reset --hard main
git cherry-pick eaf3fe75d938c14326f4f94fe027e97ec33d7ad2
```
because I think the only commit that is actually new here is the tip.

But there are conflicts. I'll take another crack at it next week. Once I resolve the conflicts, my next steps will be:

1. Verify that everything still actually works.
2. Rewrite the commit message to include all the good details written up in #32.
3. Review the actual patch, commenting on things we should follow up on later.
4. Merge it.

ctrueden:
I have now completed (1) and (2) above. @hinerm When next you are reviewing Ops PRs, this one is ready for a look!

scijava/scijava-function/templates/main/java/org/scijava/function/Inplaces.vm
```diff
@@ -74,16 +78,41 @@ public final class Inplaces {
...
+		throw new IllegalArgumentException("No Inplace of arity " + arity +
+			" and mutable index " + mutableIndex);
+	}
+
+	/**
+	 * @param c the {@link Class} of unknown arity
+	 * @return the arity of {@code c}, or {@code -1} if {@code c} is <b>not</b> a
+	 *         {@code Inplace}.
+	 */
+	public static InplaceInfo arityOf(Class<?> c) {
```
hinerm, line 112: Why does `InplaceInfo#arityOf` return an `InplaceInfo` instead of an `int`, as the other two interfaces do?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
```diff
@@ -657,6 +657,9 @@ private synchronized void initOpDirectory() {
 		List<OpInfoGenerator> generators = infoGenerators();
 		discoverers.stream().flatMap(d -> d.discover(Op.class).stream()).forEach(o -> registerOpsFrom(o, generators));
 		discoverers.stream().flatMap(d -> d.discover(OpCollection.class).stream()).forEach(o -> registerOpsFrom(o, generators));
+		Set<OpInfo> infos = opDirectory.values().stream().flatMap(Collection::stream).map(info -> opsFromObject(info, generators)).flatMap(Collection::stream).collect(
```
hinerm, line 660: you should be able to just do `flatMap(Collection::stream).forEach(info -> addToOpIndex.accept(info, log));` here and skip building the set.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/reduce/FunctionReducer.java
```diff
@@ -0,0 +1,38 @@
...
+import org.scijava.ops.engine.OpUtils;
+import org.scijava.ops.api.features.BaseOpHints;
+import org.scijava.types.Types;
+
+public class FunctionReducer implements InfoReducer{
+
+	@Override
+	public boolean canReduce(OpInfo info) {
+		boolean isFunction = Functions.isFunction(OpUtils.findFunctionalInterface(Types.raw(info.opType())));
+		boolean canReduce = info.declaredHints().containsNone(
```
hinerm: Why isn't this check in `ComputerReducer#canReduce`? Can the two classes extend an abstract base class to consolidate the routines? It looks like they are identical except the `xxxx.isXxxxx` and the two `arityOf` calls. So create an abstract superclass with unimplemented abstract methods delegating those three pieces?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/reduce/ReducedOpInfo.java
```diff
@@ -0,0 +1,222 @@
...
+		return reducedOpType;
+	}
+
+	@Override
+	public Struct struct() {
+		return struct;
+	}
+
+	@Override public Hints declaredHints() {
+		return srcInfo().declaredHints().plus(BaseOpHints.Reduction.FORBIDDEN);
```
hinerm: why not cache this, if necessary, in the constructor? Or lazily if you prefer?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/reduce/ReducedOpInfo.java
```diff
@@ -0,0 +1,222 @@
...
+
+	@Override
+	public double priority() {
+		return srcInfo.priority();
+	}
+
+	@Override
+	public String implementationName() {
+		// TODO: improve this name
+		return srcInfo.implementationName() + "Reduction" + paramsReduced;
```
hinerm, line 147: Could a space or dash between parts of the name help here?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/reduce/ReductionUtils.java
```diff
@@ -0,0 +1,315 @@
...
+	 */
+	private static String createFunctionalMethod(ReducedOpInfo info) {
+		StringBuilder sb = new StringBuilder();
+
+		// determine the name of the functional method
+		Class<?> fIface = OpUtils.findFunctionalInterface(Types.raw(info.opType()));
+		Method m = InterfaceInference.singularAbstractMethod(fIface);
+		Class<?> srcFIface = OpUtils.findFunctionalInterface(Types.raw(info.srcInfo().opType()));
+		Method srcM = InterfaceInference.singularAbstractMethod(srcFIface);
+//		Class<?> srcFIface = OpUtils.findFunctionalInterface(Types.raw(info.srcInfo().opType()));
```
hinerm: remove commented out code (in a separate commit if you want to preserve the history)

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/struct/OpResizingMemberParser.java
```diff
@@ -0,0 +1,63 @@
...
+	 * <li>That {@code srcStruct} is valid
+	 * <li>That there are {@code inputs.length} input {@link Member}s in
+	 * {@code srcStruct}
+	 * <li>That there is <b>one</b> output {@link Member} in {@code srcStruct}
+	 * </ol>
+	 * We should consider adding the evalutation of these assumptions
+	 * 
+	 * @param source the {@link RetypingRequest} from which we create the new
+	 *          {@link List} of {@link Member}s
+	 * @return a new {@link Struct} reflecting the simplified arguments / focused
```
hinerm: We're not actually returning a `Struct` here right?

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/reduce/BiFunctionWithOptionalReducer.java
```diff
@@ -0,0 +1,36 @@
+//package org.scijava.ops.engine.reduce;
```
hinerm: Delete this class?

scijava/scijava-ops-spi/src/main/java/org/scijava/ops/spi/Optional.java
```diff
@@ -0,0 +1,46 @@
...
+ * public class Impl implements BiFunctionWithOptional&lt;Double, Double, Double, Double&gt; {
+ * 	public Double apply(Double in1, @Optional Double in2, Double in3) {
+ * 	...
+ * 	}
+ * }
+ * </pre>
+ *
+ * is confusing and hard to read. Which parameters are optional in this case? Is
+ * it obvious that {@code in3} is optional just by looking at {@code Impl}? For
+ * this reason, it should be enforced that the annotation is only on one of the
```
hinerm, line 37: I didn't see any negative case tests, where the annotation appears in both interface and implementation - could you add one to show what errors are produced? Or point me at it if I missed it.

hinerm (changes requested):
Overall this is looking good and the tests are great to prove that everything is working!

ctrueden:
@hinerm Thanks so much for digging into this in detail!

Do you think it would make sense for you to take a crack at addressing your own comments, by pushing additional commits? They look relatively minor to me, and you already have your nose in it, whereas @gselzer would need to dust off his brain. Or if you don't feel comfortable making the changes yourself, maybe a pair programming session is in order?

Alternately: up until this point we were merging every PR in the chain "as is" as long as things fundamentally work, with corrections going onto later PRs. I have no strong preference whether we do that here as well; since the changes are minor it is unlikely to seriously hose up later rebases. But noting these changes to do later on top of a further along main branch is also fine.

hinerm:
>Do you think it would make sense for you to take a crack at addressing your own comments, by pushing additional commits? They look relatively minor to me, and you already have your nose in it, whereas @gselzer would need to dust off his brain. Or if you don't feel comfortable making the changes yourself, maybe a pair programming session is in order?

yep I can just make the changes @ctrueden

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
hinerm: Can't do that since we're iterating over the directory that we'll be adding to.

hinerm (approved):

PR #55, opened 2022-01-21
Move `OpUtils` back into SciJava Ops Engine

gselzer:
d521c62 moved `OpUtils` into SciJava Ops API because some API classes (`OpInfo` and `OpCandidate`, to name a few) used its functionality.

This PR moves `OpUtils` back into SciJava Ops Engine. This was made possible by the migration of relevant code into the API classes needing it. For example, `OpUtils.inputs(Struct struct)` became `OpInfo.inputs()`. With these migrations, no code in API depended on `OpUtils`, allowing it to move back into SciJava Ops Engine

Closes scijava/scijava#70

PR #59, opened 2022-04-15
Restore Package Rules

### LINKED ISSUE #92: Expose `DefaultOpEnvironment`
As we migrate away from the `OpService` paradigm and remove the dependency of SciJava Ops Engine on SciJava Common, we need a way to expose the default `OpEnvironment` implementation. While we might return to the idea of `Service`s in SciJava 3, or use some analogue to `Service`s, we should (read: have to) wrap up our `OpEnvironment` implementation in a different module, lest we reintroduce another dependency like SJC. For that reason, and to allow users to use Ops without a "`Service`" mechanism, we need to expose our `OpEnvironment` implementations.

Ideally, we'd just move `DefaultOpEnvironment` (and maybe also `DefaultOpHistory`?) into `org.scijava.ops.engine` and be done with it. However, this is made a little tricky due to the fact that `DefaultOpEnvironment` uses other classes in the `org.scijava.ops.engine.impl` package (among others), which it could not use if moved. For that reason, we will have to be crafty if moving it.

gselzer:
This PR re-establishes the package rules of scijava-maven-plugin. To do this, `OpEnvironment`, `OpHistory`, `TypeReifier` and `LoggerFactory` were rewritten as `Prioritized` implementations.

We then added `Discoverer.discoverMax()` and `Discoverer.discoverMin()`, able to discoverer the highest/lowest priority implementations for any `Class` implementing `Comparable`. 

This makes it trivial to discover `DefaultOpEnvironment`. Specifically, we can use `OpEnvironment.getOpEnvironment()` to get the highest-priority `OpEnvironment`. This is similar to the call `Context.getService(OpService.class)` in SciJava Common/ImageJ Ops. 

Remaining questions:
* [ ] I made `LoggerFactory` the discoverable interface instead of `Logger` for convenience reasons (the `Logger` implementation has constructor arguments). Is there a reason to keep `LoggerFactory`? Can we remove the constructor arguments of the `Logger` implementations?

Cannot be merged until #55 is merged, as this work is built on top of that work.


gselzer:
This would close scijava/scijava#92

ctrueden:
This PR needs a `git rebase 2c271c15b43afe0b150fb5a930ceb0ac3e1bb2e2 --onto origin/main` with conflicts addressed.

gselzer:
@ctrueden @hinerm this PR has been brought up to date, unfortunately through the process of basically doing it all over again.

I'd suggest we fast-track this PR, with the fear of having to do it all over again.

For those who review, please pay attention to the API added to the `OpEnvironment`, as while it works, I'd think we could do better. The use cases are as follows:
* Some unit tests (and use cases?) will want an *empty* Op environment. The current need is the SciJava Ops Engine unit tests
* Some unit tests will want an Op environment with all available Ops. The current need is the ImageJ Ops2 unit tests, and also Fiji scripting

To restore the package rules, we need to hide `DefaultOpEnvironment`, so using the static method in `OpEnvironment` should become the standard for obtaining an `OpEnvironment`.

hinerm (approved):

PR #65, opened 2022-11-07
imagej/image-ops2 test speedups

hinerm:
Reduces imagej-ops unit test time required by a factor of three.

See https://github.com/scijava/scijava/issues/102


hinerm:
@ctrueden @gselzer this PR creates a single OpEnvironment and eliminates some unnecessary opening of images from disk before unit tests that don't need them.

On my system `imagej-ops` tests run in less than a minute now. It's not perfect but is a reasonable improvement.

imagej/imagej-ops2/src/test/java/net/imagej/ops2/AbstractOpTest.java
```diff
@@ -75,20 +75,14 @@
  */
 public abstract class AbstractOpTest{
 
-	protected static OpEnvironment ops;
+	protected static OpEnvironment ops = new DefaultOpEnvironment();
```
gselzer: Maybe add `final` here, just in case?

gselzer (approved):
Thanks so much @hinerm! These changes look good to me!

imagej/imagej-ops2/src/test/java/net/imagej/ops2/AbstractOpTest.java
ctrueden: Done in 62aa49651fa3d0fb939ba9ca7f47040268da7d40. @hinerm Do you see any problem with doing this? I am not sure, but I could imagine that pushing the data structure construction earlier, to class initialization time rather than whenever JUnit calls the `BeforeClass` methods, could potentially alter behavior? Everything does still pass for me locally though.

ctrueden (approved):
Awesome! I followed up on one comment from @gselzer, and I also fixed some Windows line endings while at it. I think this PR is good to go! Merging now (after CI checks pass).

imagej/imagej-ops2/src/test/java/net/imagej/ops2/AbstractOpTest.java
gselzer: >  I am not sure, but I could imagine that pushing the data structure construction earlier, to class initialization time rather than whenever JUnit calls the BeforeClass methods, could potentially alter behavior?

That's a good point, but since it's only in the tests, we can change it later if we find that it's an issue, no?

ctrueden: > since it's only in the tests, we can change it later if we find that it's an issue, no?

Definitely; that was my thinking as well.

PR #67, opened 2022-12-22
Op Discovery Through YAML declaration

### LINKED ISSUE #97: SciJava Ops: create external configuration mechanism
Javadoc discovery via therapi is great for dependency-free op registration, but still requires modification to "source" code. Ultimately for adoption we need a mechanism for registering ops without any changes to source or documentation.

It seems sensible to support a configuration file syntax, e.g. in yml.

This would likely be the mechanism for including libraries in other languages.

If a library has a mechanism for introspecting its available algorithms, we could (on a per-library basis) create an automated mechanism that generates a companion ops configuration yml, and it would also provide a path for manual configuration when introspection is not possible.

gselzer:
This PR introduces a new `YAMLDiscoverer` that provides the ability to describe Ops via YAML declaration. Currently, `Class`es implementing a functional interface, `final Field`s and `static Method`s can be wrapped into Ops via YAML. This is done by writing a file `ops.yaml` within the `resources` package of a project. At runtime, the `ClassLoader` obtains all of these files on the class/module path, adding them to Ops (assuming that the functionality is exposed through JPMS (?)). An example is included below:
```yaml
info:
  source: Java
  package: SciJava Ops YAML
  version: 0-SNAPSHOT

ops:
  - op:
      name: example.add
      priority: 0.0
      source: org.scijava.ops.engine.yaml.ops.YAMLClassOp
      parameters:
        - input: in1
          description: The first input
        - input: in2
          description: The second input
        - output: out
          description: The sum

  - op:
      name: example.sub
      priority: 0.0
      source: org.scijava.ops.engine.yaml.ops.YAMLMethodOp.subtract(java.lang.Double, java.lang.Double)
      type: java.util.function.BiFunction
      parameters:
        - input: in1
          description: The first input
        - input: in2
          description: The second input
        - output: out
          description: The difference

  - op:
      name: example.mul
      priority: 0.0
      source: org.scijava.ops.engine.yaml.ops.YAMLFieldOp$multiply
      parameters:
        - input: in1
          description: The first input
        - input: in2
          description: The second input
        - output: out
          description: The product
```

The `info` section declares metadata pertaining to *all* Ops in the package:
* The `source` tells us the language the functionality is written in, and allows us to later target e.g. Python
* The `package` tells us the package the Ops are described in. I'm not sure we need this, actually.
* The `version` is used for provenance, telling us the code version.

The `ops` section iterates over all `op`s. One `Op` is created for each `op` in the list.
* The `name` is the term(s) that the Op can be searched by
* The `priority` is the Op's priority
* The `source` is a term that Ops could use to get a handle on the functionality. For Java, this is either
  * A fully qualified class name
  * A fully qualified class name **plus** a string describing a static method (with parameters, to protect against overloading)
  * A fully qualified class name **plus** a field name
* The `parameters` section provides information on the parameters. We aren't using this right now, and might want to delete it.

Discussion points:
* Is it worth keeping the `info -> package` tag, and the `parameters` tags? This PR doesn't use either, but I can see us wanting to use it later (e.g. for Python functionality).
* Should this functionality be moved into a separate module? How much of it should be moved?


scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/YAMLOpInfoDiscoverer.java
```diff
@@ -0,0 +1,12 @@
+package org.scijava.ops.engine.impl;
+
+import java.util.List;
+
+import org.scijava.discovery.Discoverer;
+
+public class YAMLOpInfoDiscoverer implements Discoverer {
```
hinerm: This class isn't used at all.. is there a reason not to delete it?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/yaml/YAMLDiscoverer.java
```diff
@@ -0,0 +1,99 @@
...
+			try {
+				Map<String, Object> opData = subMap(op, "op");
+				String identifier = value(opData, "source");
+				Optional<YAMLOpInfoCreator> c = creators.stream() //
+					.filter(f -> f.canCreateFrom(source, identifier)) //
+					.findFirst();
+				if (c.isPresent()) infos.add(c.get().create(opData, version));
+			}
+			catch (Exception e) {
+				System.out.println("Skipping op");
```
hinerm: Can you provide some info about the op being skipped?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/yaml/YAMLJavaFieldInfoCreator.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.scijava.ops.api.OpInfo;
+import org.scijava.ops.engine.matcher.impl.OpFieldInfo;
+import org.scijava.ops.engine.matcher.impl.OpMethodInfo;
+
+public class YAMLJavaFieldInfoCreator implements YAMLOpInfoCreator {
+
+	@Override
+	public boolean canCreateFrom(String source, String identifier) {
+		return source.equals("Java") && identifier.indexOf('$') != -1;
```
hinerm: Using `$` as the sole indicator of a field is problematic since it also includes inner classes. You could use the case convention to distinguish the two but that's not foolproof.

I added [an inner class test](https://github.com/scijava/incubator/commit/ab48b870892238b002369ccb3a004dd2c7dd882d) to get passing.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/yaml/YAMLJavaFieldInfoCreator.java
```diff
@@ -0,0 +1,51 @@
...
+		if (yaml.containsKey("name")) {
+			names = new String[] { (String) yaml.get("name") };
+		}
+		else {
+			String namesString = (String) yaml.get("names");
+			names = namesString.split("\\s*,\\s*");
+		}
+
+		double priority = 0.0;
+		if (yaml.containsKey("priority")) {
```
hinerm: Add an abstract superclass, implemented by the `Field`/`Method`/`Class` `InfoCreator`s, that handles `name` and `priority` and delegates the `source` parsing to the concrete implementation.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/yaml/YAMLOpTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.scijava.ops.api.OpInfoGenerator;
+import org.scijava.ops.api.OpWrapper;
+import org.scijava.ops.api.features.MatchingRoutine;
+import org.scijava.ops.engine.DefaultOpEnvironment;
+
+public class YAMLOpTest {
+
+	@Test
+	public void testYAMLClass() {
+		Discoverer serviceLoading = Discoverer.using(ServiceLoader::load) //
```
hinerm, line 33: Factor out shared behavior to `setup` method

hinerm (changes requested):
>Is it worth keeping the info -> package tag, and the parameters tags?

These feel worthwhile to me. I would keep them.

>Should this functionality be moved into a separate module? How much of it should be moved?

How much of it **can** be moved? Based on my understanding from our previous discussion, I'm in favor of consolidating code in `scijava-ops-engine` instead of submodules, since those modules aren't true "extensions" and require fundamental changes to the core module.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/YAMLOpInfoDiscoverer.java
gselzer: Nope, thanks for catching that! I deleted it.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/yaml/YAMLDiscoverer.java
gselzer: @hinerm what do you think of ba70074c?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/yaml/YAMLJavaFieldInfoCreator.java
gselzer: @hinerm we *could* just another character (e.g. `_`), the `source`s  syntax that we use in the YAML is really up to us.

hinerm: >we could just another character (e.g. _), the sources syntax that we use in the YAML is really up to us.

@gselzer it's true but I would like to avoid magic characters unless we have no other choice.

hinerm:
@gselzer I added two commits to tweak the inner-class vs. method discovery. I appreciate trying to load the class in the `JavaClassYAMLOpInfoCreator`, but that created a situation where we're relying on the `Class` info creator to be discovered *before* the `Method` info creator. This felt fragile, so I [added the class load check](https://github.com/scijava/incubator/commit/dd4f33b385733e4f39a31e65c5a20759845c3e3f) to the `Method` creator as well, and [avoid loading the class](https://github.com/scijava/incubator/commit/f0f38592b533ea660c3c4bfad9d3919f2c2edaeb) if there is no ambiguity.

If you agree with those commits: could you just rewrite your WIPs? Then we'll merge!

ctrueden:
> How much of it **can** be moved? Based on my understanding from our previous discussion, I'm in favor of consolidating code in `scijava-ops-engine` instead of submodules, since those modules aren't true "extensions" and require fundamental changes to the core module.

Is the code in question the YAML handling code? If so, I propose that it might not be specific to `scijava-ops-engine`, but more about the metadata. Imagine you are in Python and would like to invoke some convenience function (invoked via scyjava) to reason about ops, but without the full heavyweight matcher. I don't have a strong opinion—such YAML handling code could maybe be written directly in Python by leaning on the `yaml` package, although I think there would be some code duplication—but it's worth a discussion I think.

gselzer:
> Is the code in question the YAML handling code? If so, I propose that it might not be specific to `scijava-ops-engine`, but more about the metadata.

There is *some* general YAML logic, but I think right now we should leave it in the engine. It's pretty intertwined with the Ops-specific logic, and I don't yet know that we have a good use case to separate it. I say YAGNI

hinerm (approved):

PR #68, opened 2022-12-28
Wrap Numpy in Ops

gselzer:
This PR is a hacky first cut at exposing Python functionality through Ops.

Fundamentally, Python access is provided through [jep](https://github.com/ninia/jep). This PR creates a single `jep.Interpreter`, configured within a utility class. This ensures that we have control over the configuration, and only have to do it once (although there is still more configuration that needs to be done.)

Ops themselves are declared in the `ops.yaml` resource file, and are discovered by an extension of the `YAMLOpInfoDiscoverer` created in #67. The discovered `PythonOpInfo`s use javassist to create a `FunctionalInterface` implementation that:
* Transfers the input data into Python
* Executes the Python functionality
* Transfers the output data from Python

What this PR does (currently):
* Shows what it might look like to wrap Python functionality in Ops
* Creates the foundations for actually wrapping it

What this PR doesn't do:
* Wrap NumPy extensively - it includes only a couple of functions as a proof-of-concept. We'd eventually want to come up with a way (Python script?) to automate the YAML generation, and also to provide a configuration mechanism (especially aliases - I think it would be nice to e.g. alias `numpy.zeros` to `create.img`, `numpy.add` to `math.add`, etc.). It might also be cool to add a `LANGUAGE` preference to `Hints`.
* Access Python cleanly - there are many bugs in this code, and it is very much a draft.
* Add any conversions - we can only call Python Ops on Python data structures, which, right now isn't very much. 

Benefits to this approach:
* It's internal to Java - Java users get access to Python!

Drawbacks to this approach: 
* We do not (currently) share/use any scyjava functionality - the converters, especially, could be useful for simplification fallbacks. It might be a good idea to port the scyjava conversion functionality to ops somewhere down the line, if we like this approach.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,7 @@
 		<module>scijava/scijava-meta</module>
 		<module>scijava/scijava-ops-api</module>
 		<module>scijava/scijava-ops-engine</module>
+		<module>scijava/scijava-ops-python</module>
 		<module>scijava/scijava-ops-spi</module>
 		<module>scijava/scijava-parse2</module>
 		<module>scijava/scijava-persist</module>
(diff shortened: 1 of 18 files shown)


ctrueden:
This branch will continue in https://github.com/scijava/scijava. For now: closed!

PR #69, opened 2023-01-12
Handle OpMethods that the SciJava Ops Engine module cannot read

gselzer:
SciJava Ops Engine employs two different pathways for turning `public static Method`s into Ops. Both of these pathways try to obtain the `Method` by loading the owning `Class`, which fails when the `org.scijava.ops.engine` module does not `require` the module containing the `Method`. This is the case for ImageJ Ops2, and for all Op-containing modules we'd like to create in the future. This escaped our attention because we hadn't *actually* created any `OpMethod`s in ImageJ Ops2 :facepalm:

Well, fortunately for us, we don't *need* to import the Class, because `Discovery` will *always* give us an instance of that `Method` anyways, *and we don't need an owner instance because `OpMethod`s must be static*. For that reason, we can modify our javassist pathway to call [`Method.invoke`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/Method.html#invoke(java.lang.Object,java.lang.Object...)) on the `Method` instance we already have.

Points of discussion:
* What do we want to do with the other pathway,  which will no longer work (unless the engine `requires` its module, which is only the case for the engine itself right now). Should we get rid of it? Should we add a Module access check to the conditional?
* How do we want to test that this is fixed? We would have to change an Op outside of the engine to test this....

TODO:
* [x] Test this...somehow
* [x] Clean up the Javassist code

### DIFF
--- a/imagej/imagej-ops2/src/main/java/net/imagej/ops2/convert/Converters.java
+++ b/imagej/imagej-ops2/src/main/java/net/imagej/ops2/convert/Converters.java
@@ -44,8 +44,8 @@ public class Converters<I, O extends Type<O>> {
 	 * @output outputII
 	 * @implNote op names='convert'
 	 */
-	public final Functions.Arity3<RandomAccessible<I>, Converter<? super I, ? super O>, O, RandomAccessible<O>> generalConverterRA = (
-			inputRA, converter, type) -> net.imglib2.converter.Converters.convert(inputRA, converter, type);
+	public final Functions.Arity3<RandomAccessible<I>, Converter<? super I, ? super O>, O, RandomAccessible<O>> generalConverterRA =
+			net.imglib2.converter.Converters::convert;
 
 	/**
 	 * @input inputII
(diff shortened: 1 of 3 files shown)


hinerm:
>Test this...somehow

@gselzer do you need to more than just a simple test Op declared in ImageJ-ops2 that hits this path?

gselzer:
> > Test this...somehow
> 
> @gselzer do you need to more than just a simple test Op declared in ImageJ-ops2 that hits this path?

I mean, we can do that, it's just an indirect test. And we'd want to find an Op where it doesn't look weird, i.e. it would look weird to make only one `OpField` of an `OpCollection` into an `OpMethod`...

gselzer:
@hinerm alright, I rewrote the `filter.gauss` Ops as methods, so we have an implicit test.

One thing that tripped me up during this rewrite is that I mistakenly used a `double` in one of the methods, instead of the object `Double`. I'm guessing that this is something others will be tripped up on when writing `OpMethod`s, because unlike with classes and fields where you have to write a generic type the `OpMethod` won't throw a compile error if you use a primitive. 

I'm going to create an issue to see if we can get around this by boxing any primitives in `OpMethod`s. This would also be a useful feature when wrapping existing methods that take primitive parameters.

gselzer:
> What do we want to do with the other pathway, which will no longer work (unless the engine requires its module, which is only the case for the engine itself right now). Should we get rid of it? Should we add a Module access check to the conditional?

My solution, for now, is to let it remain, but to guard behind a module visibility check (see 8788ba9)

hinerm (approved):

PR #70, opened 2023-01-18
Add fallback to parameterize raw types

gselzer:
This PR fixes a `ClassCastException` arising from passing raw `Classes` to matcher calls instead of parameterized types. This would allow you to say, for example,
```java
ops.op("copy").input(img).outType(Img.class).apply();
```
The fix is made by adding another `instanceof ParameterizedType` check in `Types.isAssignable`. With this check, we ensure that that `toResolved` is also a `ParameterizedType`, meaning that `Class`es can no longer get into that code block.

This PR also adds a couple of tests to ensure exception-free behavior.


hinerm (approved):
Works for me!

PR #71, opened 2023-01-19
Improve raw type support

gselzer:
Raw types can be compile-time casted into any generic type. Let's make sure we can handle that case in the matcher!

This allows you to make Op calls like
```java
ops.op("filter.mean")
  .inType(RandomAccessibleInterval.class, Shape.class, OutOfBoundsFactory.class)
  .outType(Img.class)
  .function();
```


hinerm:
@gselzer when you get a chance can you rebase this onto `main` and check if the test I added is a) reasonable and b) still failing?

gselzer:
@hinerm I rebased over main and fixed the test that you wrote.

hinerm (approved):

PR #72, opened 2023-01-21
Gradle with dynamic source generation via Kotlin scripts

elect86:
So, the first PR was more of an example of things could appear, this has received much more love and it has been sure to work as much as possible (some tests are failing though)

Compared to the previous, since we are actually generating java files and via text, it doesn't make sense using KSP, rather, everything can be encapsulated directly in the Gradle building tool, hugely simplifying things.

I notice you are checking your source in under VSC, I'm not a big expert on source generation, but I have enough knowledge to say that this is well known anti-pattern, making things complicate, fragile (I, myself, noticed that they are already not in sync, there are several differences atm) and slower (it interferes with advanced checks and caching mechanisms). 
When you dynamically generate sources, they should be interpreted as _products_ or _artefacts_

Some advantages of this PR:
- generated sources excluded from VCS
- generation is within the build tool, which make it easier to implement (no additional/third tools) and is automatically appended at the build process, benefits from all the normal caching and performances mechanisms and guarantee always perfectly synchronicity with the producer sources (one sync big error was, for example, a `genericDoubles` missing in `OpBuilderTestOps`, other less, such as different import order or spacing/newlines) and then provide them as complementary corresponding sources along your `main` or `test` directory sources
- full code IDE support in the generation scripts, such as code highlighting, completion, navigation, refracting, searching, docs, etc etc
- Kotlin, concise and yet statically typed language
- Gradle, modern build tool
- the "templates" now incorporate everything in one single file (no more `.list` and `.vm`)

The last four points, in particular, provides a boost in terms of readability and clearness, decreasing error and bug-proneness

Since I've been playing with source generation myself in the last times, I provided this PR with the following details:
- `Globals.list` became essentially `Generator`
- every generation class script extends this `Generator`, overwriting some functions, if necessary

As a "cherry on top":
- `arities` is defaulted to `0..maxArity`, scripts using a different one have to overwrite it
- `arity` is a field in `Generator`, this brings quite relevant pleasure later on because all the previous functions which required the mentioned field, now can become simply getters, for example:
````
genericsList = ```
{ arity ->
  (1..arity).stream().map{a -> "I$a"}.collect()
}
```
````
```
	@FunctionalInterface
	public interface Arity${arity}$generics.call($arity) {
```
```kotlin
val genericsList: String get() = (1..arity).map { "I$it" }
```
```kotlin
	 @FunctionalInterface
	 public interface Arity$arity$generics {
```
Variables or getters do not need `{}` as you can see, decreasing noise in the generating `String`s
There are some cases where this requires some additional code to make function accepting one argument other than `arity` work, though, but these are very few cases

- like `arity`, also `arities` is a `Generator` field, allowing to have `forEachArity { arity -> .. }` in place of `#foreach($arity in [x..$maxArity])` when `x..$maxArity` matches the arity range of the script. In the function body, `arity` will be automatically set accordingly to the i-th iteration

Generating scripts are placed in the `buildSrc` which is special/standard module in Gradle [to abstract imperative logic](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources)

Location (`packages`) scripts in `buildSrc` doesn't matter for the generation, it's only for the dev order, some scripts aren't placed so deep as their generated source in fact (eg `opsEngine/main/matcher/OpWrappers` instead of `opsEngine/main/matcher/impl/OpWrappers`)

`GenerateCode` is the task calling each script based on a couple of input, explicit for `main: Boolean`, which is supposed to specify which src type are we generating (`main` or `test`)

Each script will call the `Generator::invoke` function, which will call in turn the abstract `generate` function implemented by the implementing class/script and then create the corresponding bare file to accommodate the generate text based on the `package` value in the generated text

This implementation is simply what I come up with at first touch, it can be of course (completely) changed and shaped based on your needs/tastes

The license and the "don't edit" warning have been extracted to variables as well saved into `Generator`

The root scijava project has dependencies on all the subprojects, making easy if you want to add all of them to a third project.

To decrease confusion, each module script is named according the module name itself, instead tens of `build.gradle.kts`, we have `collections.gradle.kts` and so on

This implementation also exploits our Gradle catalog plugin, which is the Gradle counterpart to scijava Maven poms (sync'ed with 33.2.0). It also uses a small catalog owned by the project itself in `scijava/settings.gradle.kts`.
This requires further tuning though (therapi, for example, needs an additional dependence for the processor, which doesn't look present in the scijava pom)
This allows to have a centralized place for all the dependencies _and_ provide statically typed accessors for the dependencies themselves, ie:
```kotlin
// `javassist` is a `public Provider<MinimalExternalModuleDependency>` type
implementation(misc.javassist) 
// vs
implementation("org.javassist:javassist:3.28.0-GA")
```

This also uses a Gradle incubating feature, called `TYPESAFE_PROJECT_ACCESSORS`, which offers accessors for subprojects
```kotlin
// scijavaCollections is a `public class ScijavaCollectionsProjectDependency extends DelegatingProjectDependency` type
implementation(projects.scijavaCollections)
// vs
implementation("org.scijava:scijava-collections")
```

For dependencies which aren't proper modules, I'm using `extra-java-module-info` [plugin](https://github.com/gradlex-org/extra-java-module-info/) to fill the void

The same concept of root dependencies connection is applied to the root `test` task and the corresponding tasks of the subproject, if you run `./gradlew test` it will call in turn all the tests of the underlying modules, apparently failing test in one module doesn't let following modules to run their own test, that's why `discovery-therapi` (9/9 failed), `ops-engine` (96/1542 failed) and `persist` (2/2 failed) are commented out (disabled) in `scijava/build.gradle.kts`. 
If you want to run some specific module tests only, `/gradlew :scijava:scijava-common3:test`

`assemble` task only build the given project, `build` will build **and** execute all the tests
You can either use Idea gui or the cli, if you have any questions, doubts and so on don't hesitate, even short and direct questions on Zulip

Although I applied some of the same changes to imagej as well, I focused primarily on scijava, being the latter a direct dependency of the first one.

If this goes well, I'll finish up imagej as well

The original generated classes are still there, I just removed the `.java` extensions to avoid interferences 

Ps: I tested and developed everything with jdk11
Pps: I postponed the `module-info.java` generation file as well for the moment


ctrueden:
The scijava-ops codebase has migrated to https://github.com/scijava/scijava. So that this work is not lost, I have migrated it to the [`gradle-and-kotlin-gen` branch of scijava](https://github.com/scijava/scijava/compare/stale/gradle-and-kotlin-gen).

elect86:
If you are still interested in this, I can reboot it once more time

PR #74, opened 2023-03-14
Refactor OpBuilder to improve usability

hinerm:
The main API changes include:
* A new `arityN()` [step in the builder](https://github.com/scijava/incubator/blob/53ae8a675562f8eb60c3472414658011804c132d/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java#L108), after specifying Op name but before input value/type(s). This avoids the `arity * inputMethods` method proliferation.
* The `matchInplace` static utility methods [now include two digits](https://github.com/scijava/incubator/blob/53ae8a675562f8eb60c3472414658011804c132d/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java#L17527), indicating both the arity and the mutated input. This avoids having to try and count the desired number of total inputs when selecting between overloaded `mutateInplace1` methods.

All public API should now have javadoc. It should be fairly consistent across similar methods. Some particular areas of attention:
* The [intro and examples](https://github.com/scijava/incubator/blob/53ae8a675562f8eb60c3472414658011804c132d/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java#L48-L96)
* How [hints](https://github.com/scijava/incubator/blob/53ae8a675562f8eb60c3472414658011804c132d/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java#L321) overloads are javadoc'd
* How [inplace and mutate](https://github.com/scijava/incubator/blob/53ae8a675562f8eb60c3472414658011804c132d/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java#L1798-L1799) are documented: non-`inplace`/`mutate` methods just `@see` to the first arity, while each individual `inplace`/`mutate` method `@see`'s to the complementary arity `mutate`/`inplace`. Alternatively we could link to ALL the other arities in one or both of these location(s). Or we could make a single dump of all methods in the class javadoc... but given the nature of the builders I think the method javadoc is much more visible.

Other thoughts:
I was considering expanding the `[I/O][T/V/U]` acronyms in some abbreviated way. I initially found them confusing and didn't like the "single-letter differentiation". After looking at the builder for a while now I'm pretty used to them, and hopefully the javadoc is clear what decision you're making at each point. So I decided to just [define the acronyms](https://github.com/scijava/incubator/blob/53ae8a675562f8eb60c3472414658011804c132d/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java#L74-L76) in the class javadoc.

Currently a draft because of the API changes. @gselzer @ctrueden if you're ok with these changes I will update the builder uses throughout the incubator. No rush; just let me know!


scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
```diff
@@ -229,10 +248,22 @@ public Type getType() {
 				Object.class, hints));
 		}
 
+		/**
+		 * Directly run this type-unsafe {@link org.scijava.function.Producer} op and get its output.
```
ctrueden: I would suggest language like "Match and then immediately run...", or "Directly match and run..." because the thing being built by the op builder mechanism is not an op itself. It's more like a query being built up.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
```diff
@@ -191,18 +191,32 @@ private void checkRefs(int mutableIndex, String label, Object... objects) {
...
 
+		/**
+		 * Match an op using the indicated output class.
+		 */
 		public <O> Arity0_OT<O> outType(final Class<O> outType) {
 			return outType(Nil.of(outType));
 		}
 
+		/**
+		 * Match an op using the indicated output class, with generic parameter(s) preserved.
```
ctrueden: This one isn't a class, it's a generic type, specified using the `Nil` mechanism.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
```diff
@@ -191,18 +191,32 @@ private void checkRefs(int mutableIndex, String label, Object... objects) {
 	 */
 	public final class Arity0 {
 
+		/**
+		 * Match a {@link org.scijava.function.Computers} op that can use the given pre-allocated output instance.
```
ctrueden: I'm not sure about linking to the `Computers` container class here. It's true that the vast majority of the time, the reason to pass an output instance is because you are going to request a computer op of some arity. But not *necessarily*. It could/should match any op with a `@Container` parameter in its functional interface, I think.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
```diff
@@ -276,22 +313,38 @@ public Type getType() {
...
+		 *
+		 * @see #producer
+		 * @return The {@code O} created by this op
+		 */
+		public O create() {
+			return producer().create();
+		}
+
+		/**
+		 * As {@link #create} but using the provided {@code Hints}.
```
ctrueden: These `Hints` signatures weren't part of my original design, and I think we can do better. We could have a separate `hints(Hints)` call in the chain that associates the hints into the build, so that we don't have to propagate twice as many method signatures over all the builder classes. Or it could maybe just go into the original `op(String)` function, as a second `op(String, Hints)` right away. What do you think?

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
```diff
@@ -308,18 +361,32 @@ public Arity0_OV(final O out) {
 			this.out = out;
 		}
 
+		/**
+		 * @return An instance of the matched {@link org.scijava.function.Computers} op, e.g. for reuse.
```
ctrueden: This description implies that a match has already taken place. In actuality, every call to this method will do another match action. I would change the javadoc accordingly to say "Match and return a computer op" or some such.

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -133,18 +133,50 @@ public class OpBuilder {
 	 */
 	public final class Arity0 {
 
+		/**
+		 * Match a {@link org.scijava.function.Computers} op that can use the given pre-allocated output instance.
```
ctrueden: It's misleading to suggest that calling this method performs a match. It's building a request. The match happens later. Instead of "Match" I would say "Specifies" or "Requests" or "Indicates" or some such.

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -433,28 +433,38 @@ public class OpBuilder {
 			this.outType = outType;
 		}
 
+		/**
+		 * Match a {@link org.scijava.function.Functions} op to produce an instance of this builder's {@code outType}.
+		 *
+		 * @see #computer
```
ctrueden: Why is there a link to `#computer` here? For contrast with `function`? I think you can add a description after the `@see #computer` blurb explaining *why* the reader should see it, which could be helpful here.

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -433,28 +433,38 @@ public class OpBuilder {
...
+		 * As {@link #function}, but match using the provided {@code Hints}.
+		 */
 		public $functionArity.call($arity)$generics.call($arity) function(final Hints hints) {
 			return matchFunction(env, opName, $inputTypesWithOutput.call($arity), hints);
 		}
 
+		/**
+		 * Match a {@link org.scijava.function.Computers} op to work with pre-allocated output.
+		 *
+		 * @see #function
```
ctrueden: Same here: would be good to explain that the link to `#function` is by contrast/alternative, a different style of invocation than this one or some such. Also, why not also `@see #inplace` in that case?

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -652,56 +652,140 @@ public class OpBuilder {
 #end
 		}
 
+		/**
+		 * Match a {@link org.scijava.function.Computers} op that can use the given pre-allocated output instance.
+		 *
```
ctrueden: There are even more `@see` blocks here, and no explanation why...

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -47,13 +47,52 @@ import org.scijava.types.Types;
 
 /**
  * Convenience class for looking up and/or executing ops using a builder
- * pattern.
+ * pattern. The general order of specification:
+ * <ol>
+ *   <li>The op name</li>
+ *   <li>The number of input(s)</li>
```
ctrueden: I love the additional level of detail here! But it gives me an idea to make the builder more succinct: the new `arityX` methods given subsequent to `op(opName)` do not take any arguments, right? So what if we got rid of `op(String)` in favor of directly asking for the arity from the `OpEnvironment`? So we'd have something like:

```java
ops.unary("math.sqrt").input(5).outType(Double.class).apply()
ops.binary("math.pow").inputs(6, 3).output(myDoubleTypeResult).compute()
ops.arity3("logic.ternary").inputs(myConditionalValue, myTrueValue, myFalseValue).apply()
```
and so forth.

@hinerm What do you think?

ctrueden: Relatedly: this javadoc would be more accessible if put onto the `.op(String)` method of the `OpEnvironment`, don't you think? That's where people will be discovering things via autocomplete. Is it likely they are going to stumble upon the javadoc for the `OpBuilder` class itself, when it is usually not imported and referenced directly?

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -47,13 +47,52 @@ import org.scijava.types.Types;
...
+ *     </ul>
+ *   </li>
+ * </ol>
+ * The first two steps are required, at a minimum. The choices you make will
+ * determine the <i>type</i> of Op that is matched:
+ * <ul>
+ *   <li>No inputs &rarr; <code>Producer</code> or <code>Computer</code></li>
+ *   <li>Inputs with an output <i>value</i> &rarr; <code>Computer</code></li>
+ *   <li>Inputs with an output <i>type</i> &rarr; <code>Computer</code> or <code>Function</code></li>
+ *   <li>Inputs with no output &rarr; <code>Function</code> or <code>Inplace</code></li>
```
ctrueden: Maybe change this line to read "Inputs with no output -> Inplace or return-type-unknown Function"?

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -47,13 +47,52 @@ import org.scijava.types.Types;
...
+ * </ul>
+ * <br/>
  * <p>
- * TODO: Examples
+ * Note that the intermediate builder steps use the following acronyms:
```
ctrueden, line 53: OK, so this part I would keep here on the `OpBuilder` javadoc, and not move anywhere more visible. The IV/OV et al acronyms were always intended to be as hidden as possible. Obviously they are public classes, and the IDE will report the return types of things, but that's exactly why I made them as short as possible: because I didn't want people worrying about these types—instead the goal is to focus them on the sequence of chained method calls, and what methods become available to use at each step. The return types are an implementation detail. All of that said, I think the explanation here is nice for the people who are curious.

ctrueden (commented):
@hinerm Thanks a lot for working on this! I wrote my thoughts as I went through the commits. Aside from tweaks to javadoc in various places, my big question is whether we should collapse the `op(String)` and `arityX()` calls into one layer. A secondary implied question is whether to use the terms `nullary`, `unary`, and `binary` instead of `arity0`, `arity1`, and `arity2` for those methods.

P.S. I force-pushed the branch to have one more leading commit, which I've already pushed to main: an update to pom-scijava 34.1.0. This eliminates 12500 lines of build log from the repeatedly echoed active profiles.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
hinerm: @ctrueden the `Arity0_OV` intermediate builder returned here only has [compute() and computer()](https://github.com/scijava/incubator/blob/f77076b5423522257ec3d7cc2f2e3d57e41c38e6/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java#L433-L456) methods. @gselzer does this seem incorrect to you?

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
hinerm: I definitely am in favor of a second `op(String, Hints)` signature! I'll make the change

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
hinerm: I think it's important to have the `@return` on the terminal builder steps, when applicable. I agree with the correctness of your statement, but was trying to avoid repetition in description and return. I'm thinking maybe something like this will work?
```
/**
 * Query an Op based on the choices made with this builder.
 *
 * @return An instance of the requested Op, if matched.
 */
```

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
hinerm: Yeah I want each intermediate builder choice to link to each other option so that the developer always has a guide to what all their options are. I agree blurbs would be nice - I'll get on that!

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
hinerm: Will resolve all these as indicated in https://github.com/scijava/incubator/pull/74#discussion_r1136486448

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
hinerm: > Also, why not also `@see #inplace` in that case?

Note that this is in the `IT_OT` builders which do not have `inplace` methods. @gselzer @ctrueden is this a mistake? It seemed intentional to me that only the `IV_*` builders had `inplaces` but if the `IT_*`'s should too I'm happy to add them.

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
hinerm: > Relatedly: this javadoc would be more accessible if put onto the `.op(String)` method of the `OpEnvironment`, don't you think?

100% agree. Will add links to `OpEnvironment` if it makes sense after the following discussion.

> So what if we got rid of `op(String)` in favor of directly asking for the arity from the `OpEnvironment`?

My first reaction is that I love this idea. Caveat thoughts:
* I like the elegance of `unary`, `binary`, etc.. but I also like the consistency of `arity0...N`. I can see arguments for and against both in an autocomplete environment (`unary`/`binary` could get lost in the `arity` swarm.. but if you know which you want you can get there with fewer characters).. so I am torn.
* Note that there would be two entries for each arity, one with `Hints`.. maybe the `Hints` should be applied at the very end, since `apply`/`compute`/etc... don't currently take any parameters? That will limit the total number of methods generated and feels consistent with the earlier logic for moving the `Hints`. It would also keep the `OpEnvironment` API cleaner.
* Would `OpEnvironment` then be auto-generated with the different arity methods? If so, do we even want two separate classes? If not, then are we conflating concerns between the two classes?

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
ctrueden: What happens if no op can be matched? An exception, right? So maybe you could just write something like `@return An instance of the op matching the request` and then also have a `@throws` to describe the situation where it fails?

ctrueden: And I agree about the desirability of having `@return` declared. I didn't mean to suggest to get rid of it, technically. We should find a way to keep it while minimizing redundancy as you say.

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
ctrueden: About the lack of `@see #inplace`: I see, that makes sense not to have it then, since this is in a part of the subtree not aiming toward an inplace op match. :+1: I doubt it's a mistake—when designing these paths I considered very carefully which paths would make sense for each type of op.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
ctrueden: I understand that as written, this builder path is aiming toward producing only computers, not functions nor inplaces. However, I'm wondering if it might make sense for it to produce something else, of some custom functional interface compatible with the specified constraints? We don't have a matching endpoint for this currently, but one could imagine something like `.match(Class<?> customFunctionalInterface)`. Of course, if we ever were to add that, we could change the javadoc accordingly at that time, I suppose.

scijava/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
ctrueden: @hinerm Some thoughts in response to your thoughts:

* We could put both `unary`/`binary` and `arity1`/`arity2` so they're easier to find? Or if the redundancy is too yucky, I guess I lean toward `unary`/`binary`; we already special-case the `Function` and `BiFunction` because they already exist and creating `Functions.Arity1` and `Functions.Arity2` felt redundant, so in that way, `unary`/`binary` is more consistent naming-wise.
* For the `Hints` support... I think wanting to pass a `Hints` object is in the small minority of cases. So what if we *kept* `op(String)` and `op(String, Hints)` signatures, but *also* added `unary(String)` as a default interface method that calls `op(String).arity1()` under the hood, and `binary(String)` calling `op(String).arity2()`, etc.? Then people have the nicest + shortest + most readable way, but also the way to specify hints that is still pretty elegant and nice to read.
* For the auto-generation: my vote would be to just add the 16-17 needed default method one-liners by hand and not worry about autogenerating them. If we ever change the number of built-in arities, we can just update the `OpEnvironment` manually as well.

ctrueden: One other way of handling the hints would be to change all the various builder classes to extend a common `OpBuilderStep<T extends OpBuilderStep<T>>` base class, which has a `public T hints(Hints hints)` method that stores the hints into a protected hints variable, then returns itself. So then you could chain-add your `Hints` object at *any* step of the build.

ctrueden: I'm liking the use of `unary` and `binary` from the `OpEnvironment`, while keeping it `arity1()` and `arity2()` from the `OpBuilder`, more and more as it settles in my brain. We could even cross-link them in the javadoc, where the javadoc for `arity1()` says something like `@see OpEnvironment#unary for a shorthand way to access this path` and similar for binary.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
hinerm: I changed it anyway since it's not performing the match at this point

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
gselzer: I also like `op(String, Hints)`

gselzer:
@hinerm thank you for making these changes. 

I really appreciate the work you did on the javadoc. That was needed.

I still do not see the value in adding these `arityX` methods to the `OpBuilder` syntax. While the implementation seems fine, I still believe that designing to improve auto-complete is misguided. is If I'm outnumbered, then so be it, but I think that syntax like this is a step backward.

I feel like for a long time I've been saying that I don't see the value in autocomplete in Ops, hence not bothering to fix it, so maybe I'll try to spill out my thinking below:

My intuition says that **if you know a function that does what you want to do, then you should just call that function directly**. I see Ops as a way to help people who *don't know what function to call*. By using Ops to call a specific function, you'll take a performance hit due to matching, parameter conversion, etc. 

So, if you're using Ops to find some unknown function matching your needs, I only see autocomplete as being useful if it can tell you:

1.  What Ops are in the environment.
2.  How to call any given Op in the environment (i.e number of parameters and types of each).

The OpBuilder syntax can't do either of these things.

With that said, what information do I gain from autocomplete? The only thing that I can think of is that you can see that for the `inputs` method you need to pass Objects, while for the `inTypes` method you need to pass all `Class`es or all `Nil`s. I admit this is cleaner with your syntax, but I'd argue that *it should also be clear from the method names*. If it's not, we need to change the method names or people will have trouble with Ops when they don't have autocomplete.

Again using my intuition above, I really envision Ops being most useful when you *don't* have these smart IDE features telling you about the functions in your environment, like with PyImageJ or in Fiji's Script Editor. The OpBuilder is a solution to this problem, promising you that it will try its best to find a function you can use to fulfill your task. Writing `op(...).input(...)...` *every time* is really elegant, and aligns well with (at least my) natural thought processes. Adding `arityX` makes me consider how many inputs I'll be providing before I enumerate them in the Op call, which increases my mental burden.

At the end of the day I'm probably bikeshedding, hence my approval if you guys really like this change, but I still think that we are focusing on the wrong place here.

hinerm:
@gselzer 

> I see Ops as a way to help people who _don't know what function to call_

Ah interesting. Amusingly this is _not_ the same as how I see Ops. 😆 So not surprising that there are different opinions here.

I would expect Ops to be used when you know _the algorithm_ you want, but don't know what _the inputs_ will be (and/or want to future-proof for unknown future inputs). The cost of time and knowledge complexity is paid for by the fact that core code (calling the algorithms) doesn't have to change for future use-cases to evolve.

Based on [past discussions](https://github.com/scijava/scijava/issues/111), I don't know that the `OpBuilder` will actually have long-term use. I think users will want some GUI built on top of some improved `Help` functionality. Developers will [use python](https://github.com/scijava/scijava/issues/113) so they have namespaces.

>I really envision Ops being most useful when you don't have these smart IDE features telling you about the functions in your environment, like with PyImageJ or in Fiji's Script Editor

I think these use cases will be great and will use the enhanced pythonic dynamic entry points - but those things are in some obfuscated future. I don't think they will use the `OpBuilder`. But **right now** I see the `OpBuilder` as "the way to use Ops", and at the very least as the entry point for new developers to show what Ops can do and build buy-in.

So with that in mind, the `OpBuilder` is essentially a series of decisions. More decisions/choices = more cognitive load. Combining `arity` and input type into one step gives you a number of choices equal to the **product** of each individually, whereas splitting them into two steps turns the total choices into a **sum**.

I don't want to portray this as "just for auto-complete," although I know that's how I first brought it up [in this issue](https://github.com/scijava/scijava/issues/111). Auto-complete just illustrated how many things the user has to choose between. The reality was, when I first was using the `OpBuilder`, it was in an IDE and I couldn't actually tell there were 3 different input choices (class/Nil/value) and I don't want other developers to bounce off things like this.

>Adding arityX makes me consider how many inputs I'll be providing before I enumerate them in the Op call, which increases my mental burden.

This is totally valid. I feel the same way. I don't like having a step to select arity. I don't *want* to think about arity at all. I *want* an `Ops.run(blah)` to just give it my stuff and have it do the right thing. This is exactly why I have reservations about `ops.unary("math.pow")`, `ops.binary("math.add")`, etc... but it seems like the cost of type-safety.

Different developers will probably find one option fits them better than the other. Having 2 input steps feels more comfortable to me, but ultimately if the `OpBuilder` sticks around I think we should show both options to a test group and just use whatever is the least offensive.

ctrueden:
@gselzer wrote:
> My intuition says that if you know a function that does what you want to do, then you should just call that function directly.

The downside to calling a function directly is that you lose the extensibility provided by Ops. You will no longer gain potential performance benefits for more specialized algorithm implementations. I agree with @hinerm's point here:
> I would expect Ops to be used when you know the algorithm you want, but don't know what the inputs will be (and/or want to future-proof for unknown future inputs).

With emphasis on the "want to future-proof for unknown future inputs" and especially keeping the door open for more performance-optimized implementations, as highlighted in [Ops design goal 3: "Very fast"](https://imagej.net/libs/imagej-ops/#design-goals).

@hinerm wrote:
> Based on past discussions, I don't know that the OpBuilder will actually have long-term use. I think users will want some GUI built on top of some improved Help functionality. Developers will use Python so they have namespaces.

Here I disagree; I think the niche of the op builder is for developers programming in Java, who want type safety. There will continue to be some people programming in Java.

> Combining arity and input type into one step gives you a number of choices equal to the product of each individually, whereas splitting them into two steps turns the total choices into a sum.

Indeed. I think this is a nice way to summarize the benefit of splitting out the arity layer.

> The reality was, when I first was using the OpBuilder, it was in an IDE and I couldn't actually tell there were 3 different input choices (class/Nil/value) and I don't want other developers to bounce off things like this.

:point_up: To me, this is the most crucial point. @gselzer and I cannot use the builder from a new developer's perspective, whereas @hinerm had that window of opportunity, and observed a point of confusion, and identified a way to relieve it. Therefore, I accept that the change is valuable.

hinerm:
@ctrueden @gselzer alright I think I addressed all your feedback. Let me know if there's anything else you'd like to see! Thanks!

ctrueden:
Thanks again for all your work on this, @hinerm!

Earlier, we had discussed potentially compacting the `op(...)` and `arityX()` calls into one step. Did we ultimately decide against that? Because I still think it would be slick to do that:

```java
ops.arity3("logic.ternary").input(condition, trueValue, falseValue).apply()
ops.binary("math.add", hints).input(img1, img2).output(result).compute()
```
etc.

What is the benefit of keeping the `op(...)` step separate? One reason was that it provides a central place to pass the `Hints` if you have some... anything else? We could also pass hints like:
```java
ops.binary("math.add").hints(hints).input(img1, img2).output(result).compute()
```
so that the `OpEnvironment`'s `unary`/`binary`/`arity3`/etc. methods don't get doubled signature-wise.

@hinerm If you *like* the idea of collapsing the `op` and `arityX` methods into one level, but are sick of doing work on this branch, then I volunteer to do it, since it would make me happier.

Alternately: we could loop in @tpietzsch and @tinevez next week to see which form they find the best.

What do you think?

hinerm:
> Earlier, we had discussed potentially compacting the `op(...)` and `arityX()` calls into one step. Did we ultimately decide against that?

I [put them in the OpEnvironment](https://github.com/scijava/incubator/blob/12e8d342293af3236e5fb38097843c69bebe6855/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java#L231-L439)

ctrueden (approved):
LGTM :rocket:

ctrueden:
> I [put them in the OpEnvironment](https://github.com/scijava/incubator/blob/12e8d342293af3236e5fb38097843c69bebe6855/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java#L231-L439)

Awesome, sorry I missed that! I saw that the builder usages across the codebase were still using the long-form `op(...).arityX()` style, so didn't realize it was included.

hinerm:
> One reason was that it provides a central place to pass the `Hints` if you have some... anything else? We could also pass hints like:

I think we might want to do another separate pass on `Hints`.. I'm not confident in how they're used right now. Maybe we can/should discuss in another issue...

hinerm:
> Awesome, sorry I missed that! I saw that the builder usages across the codebase were still using the long-form `op(...).arityX()` style, so didn't realize it was included.

yeah I did the changes in a dumb order 🙄

PR #78, opened 2023-06-20
Use LambdaMetafactory for OpMethods with dependencies

gselzer:
This PR enables `OpMethod`s **with dependencies** to be wrapped into functional interfaces via the `LambdaMetafactory`, significantly improving performance (see [here](https://forum.image.sc/t/performance-profiling-ways-of-invoking-a-method-dynamically/27471?u=gselzer) for more info on the performance of `LambdaMetaFactory` over dynamic invocation, which is what the other pathway does).

To enable `LambdaMetaFactory` to curry dependencies, they **must precede any other parameters in the method** - therefore, this PR also updates the `MethodOpDependencyMemberParser` to enforce that any dependencies come first.


scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/OpMethodDependencyPositionTest.java
```diff
@@ -17,82 +21,93 @@
...
-		List<Long> expected = Arrays.asList(1l);
-		assertIterationsEqual(expected, out);
+	public void testOpDependencyBefore() throws NoSuchMethodException {
+		var m = this.getClass().getDeclaredMethod(//
+			"goodDep", //
+			Function.class, //
+			List.class, //
+			List.class //
+		);
+		var info = new OpMethodInfo( //
```
hinerm, line 42: I don't really like seeing `OpMethodInfo` manually constructed outside of a test to explicitly make sure `OpMethodInfo` construction works as intended. My intuition is that ops tests should go through an `OpCollection` as they would be used in production.

hinerm (changes requested):
I decided to try adding an `OpMethod` with a dependency in a non-scijava package, e.g.:
```
	@OpMethod(names="test.ijOpWithDep", type=java.util.function.Function.class)
	public static <T> Double opWithDep(
			@OpDependency(name="stats.mean")Computers.Arity1<Iterable<T>, DoubleType> meanOp,
			Double in) {
				System.out.println(meanOp);
		return in + 1;
	}

	@BeforeAll
	public static void addNeededOps() {
		ops.register(new CopyIITest());
	}

		@Test
	public void testAthing() {
		Double v = ops.op("test.ijOpWithDep").arity1().input(2.0).outType(Double.class).apply();
		System.out.println(v);
	}
```

which I just attached arbitrarily to the `CopyIITest`. This produced the error: "OpMatching No MatchingRoutine was able to produce a match". I assume because it lacks read access to the package?

Shouldn't it be falling back to reflection? Or is there something wrong in my syntax?

Will this be resolved by https://github.com/scijava/scijava/issues/121, or not because it's Javassisting vs. using an existing op? I believe you have a fix for that issue in progress right?

What I'd like to see in this PR is tests for `OpMethods` with `OpDependencies` in non-scijava components working, potentially after rebasing onto the fix for https://github.com/scijava/scijava/issues/121 (or whatever is necessary to make this work).

As a side note, why is this necessary when I do `OpMethod` annotations?
```
	@BeforeAll
	public static void addNeededOps() {
		ops.register(new CopyIITest());
	}
```

gselzer:
> This produced the error: "OpMatching No MatchingRoutine was able to produce a match". I assume because it lacks read access to the package?

Nope, it's because your dependency is incredibly difficult to match - you'd be hard-pressed to find a `mean` Op that can operate on an unbounded `T`!

Changing to the following works *with `LambdaMetaFactory`*:

```java
	/**
	 * @implNote op names='test.ijOpWithDep', type='java.util.function.Function'
	 */
	public static <T> Double opWithDep(
			@OpDependency(name="stats.mean") Function<Iterable<Double>, Double> meanOp,
			Double in) {
		System.out.println(meanOp);
		return in + 1;
	}


	@Test
	public void testAthing() {
		Double v = ops.op("test.ijOpWithDep").arity1().input(2.0).outType(Double.class).apply();
		System.out.println(v);
	}

```



> What I'd like to see in this PR is tests for `OpMethods` with `OpDependencies` in non-scijava components working, potentially after rebasing onto the fix for [scijava/scijava#121](https://github.com/scijava/scijava/issues/121) (or whatever is necessary to make this work).

I wrote 46f9905 to implicitly test this - let me know what you think.

> As a side note, why is this necessary when I do `OpMethod` annotations?
> 
> ```
> 	@BeforeAll
> 	public static void addNeededOps() {
> 		ops.register(new CopyIITest());
> 	}
> ```

This is necessary for tests in the engine because, to speed up the tests, we don't discover anything by default, so you have to tell the `OpEnvironment` the list of sources that should be used to generate Ops. This is *not* needed in ImageJ Ops2's tests, because we just discover everything.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/OpMethodDependencyPositionTest.java
gselzer: > I don't really like seeing `OpMethodInfo` manually constructed outside of a test to explicitly make sure `OpMethodInfo` construction works as intended.

Well, it is *in* a test :sweat_smile: 

> My intuition is that ops tests should go through an `OpCollection` as they would be used in production.

If we were testing an actual Op that would be used in production, I'd agree, but the thing we're trying to test here (i.e. ensuring that the `OpMethodInfo` is invalid *because* of the position of the `OpDependency`) would be impossible to test if we went through the `OpBuilder` in some way. This is also the minimal amount of code needed to test this behavior - i.e. it's a unit test.

For the above reasons, I believe that it should be left the way it is. I suppose we could refactor this into a test that ensures improper Ops are not matched, but I think that way of writing tests is much more prone to error.

With that said, we generally *don't* test like this, and there is something to be said for consistency. If you really think consistency is important, feel free to rewrite the test like I described, or in your own preferred way!

hinerm: @gselzer It is fine to have a focused unit test for `OpInfo` creation but it is not sufficient. There also need to be framework-level tests that show how invalid ops are handled.

I added my thoughts after trying to tweak these tests to https://github.com/scijava/scijava/issues/25. So, it's fine to wait until that issue is addressed to make more changes.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/OpMethodDependencyPositionTest.java
```diff
@@ -17,82 +21,93 @@
...
+				m, //
+				Computers.Arity1.class, //
+				new DefaultHints(), //
+				"test.dependencyBeforeAndAfterInput" //
+		);
+		ValidityException exc = info.getValidityException();
+		String expMsg = "java.lang.IllegalArgumentException: Op Dependencies in " +
+				"static methods must come before any other parameters!";
+		Assertions.assertEquals(expMsg, exc.problems().get(0).getMessage());
```
hinerm, line 110: @gselzer String comparison is a fragile mechanism to tie a test to. It is an opening for this test to fail even though behavior is still correct.

To me it is a sign that something is fundamentally missing that needs to be addressed in https://github.com/scijava/scijava/issues/25 or https://github.com/scijava/scijava/issues/105 (and maybe those issues could be merged?)

hinerm (approved):

PR #80, opened 2023-07-12
Create API for global progress listeners

gselzer:
This PR adds API for introducing global progress listeners, in anticipation of end-users or external framework wanting to monitor progress without registering every Op.

Beyond the new global API, this PR also introduces additional changes needed to make `Task`s more informative and robust. A list of the changes is provided below:
* A `description` field is added to the `Task` class to make it clearer which `Task`s describe a given execution. To support Ops, `AbstractRichOp.toString()` now provides a clear, more informative name.
* To prevent multiple pings on completion, a safeguard was introduced to ensure that a `Task` completes only once.
* Task updates now take a `long` defining the number of steps, instead of an `int` - this makes the step math more uniform.

### DIFF
--- a/imagej/imagej-ops2-tutorial/src/main/java/net/imagej/ops2/tutorial/ReportingProgress.java
+++ b/imagej/imagej-ops2-tutorial/src/main/java/net/imagej/ops2/tutorial/ReportingProgress.java
@@ -119,18 +119,22 @@ public class ReportingProgress implements OpCollection {
 		// This ProgressListener simply prints out the status of the Op
 		// to the console, but we could print out something else,
 		// or pass this information somewhere else.
-		ProgressListener l = //
-			task -> System.out.printf("Op progress: %.2f\n", task.progress());
+		ProgressListener l = task -> //
+			 System.out.printf("Progress of %s: %.2f\n", task.description(), task.progress());
+		// To listen to Op progress updates, the ProgressListener must be registered
+		// through the Progress API. To listen to all Op executions, use the
+		// following call:
+		Progress.addGlobalListener(l);
+		// If listening to every Op would be overwhelming, the Progress API also
+		// allows ProgressListeners to be registered for a specific Op, using the
+		// following call:
+		// Progress.addListener(op, l);
 
 		// Get the function.
 		var op = ops.unary("tutorial.long.op") //
 			.inType(Integer.class) //
 			.outType(new Nil<List<Long>>() {}) //
 			.function();
-		// Listening to every Op would be overwhelming.
-		// Ops must be deliberately linked to a ProgressListener through the
-		// Progress API.
-		Progress.addListener(op, l);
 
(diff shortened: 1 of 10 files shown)


ctrueden (commented):
The changes to `Task` to make it more robust are great!

But I'm not convinced about the `Progress.addGlobalListener` feature. What is a concrete use case where this is needed? I can't think of one.

Furthermore, the scijava-progress concept is very general, beyond only scijava-ops. You could in theory subscribe to updates to anything, right? So then this global listener thing really only listens to *everything*, with no way to narrow it to particular scopes. I feel like, if we are going to allow listeners that are broader than individual objects, we may want to allow listening to *classes* of objects, and then if you listen to `Object` you listen to everything. But now we are starting to reinvent scijava-common's `org.scijava.event` package, which already has that functionality in its event pub/sub model. Maybe it's worth considering if/how these two subsystems could be combined?

gselzer:
> What is a concrete use case where this is needed? I can't think of one.

The original use case that I was thinking of was a Progress UI, allowing the user to keep track of Ops run in scripts and potentially also run graphically through a search bar. I don't know how you'd otherwise allow internal mechanisms to listen to progress - you'd probably need a callbacks mechanism...

The other use case here is convenience - the current API requires you to register the `ProgressListener` and the progressible `Object` before the `Object` starts providing progress updates. While the granularity is nice, there are two drawbacks on that approach:
1. You cannot track the progress of executions performed directly from the `OpBuilder` call (like the one below), because the user does not have a handle on the Op.
```java
ops.op("foo.bar").arity1().input(foo).output(bar).compute();
```
2. You must register *every* Op with the `ProgressListener`, which will probably double the lines of code in your script.

It was this second use case that I encountered when I updated the tutorials, and made me want to add this functionality.

> But now we are starting to reinvent scijava-common's `org.scijava.event` package, which already has that functionality in its event pub/sub model. Maybe it's worth considering if/how these two subsystems could be combined?

That's a good point, although I think we may still need the functionality to subscribe globally?

gselzer:
I merged these changes because I think it is an improvement in usability.

@ctrueden said:

 

> The changes to `Task` to make it more robust are great!
> 
> But I'm not convinced about the `Progress.addGlobalListener` feature. What is a concrete use case where this is needed? I can't think of one.
> 
> Furthermore, the scijava-progress concept is very general, beyond only scijava-ops. You could in theory subscribe to updates to anything, right? So then this global listener thing really only listens to _everything_, with no way to narrow it to particular scopes. I feel like, if we are going to allow listeners that are broader than individual objects, we may want to allow listening to _classes_ of objects, and then if you listen to `Object` you listen to everything. But now we are starting to reinvent scijava-common's `org.scijava.event` package, which already has that functionality in its event pub/sub model. Maybe it's worth considering if/how these two subsystems could be combined?

I think this is still a fair point, but I don't want to spend time right now thinking about it. I'd rather think about it when we have a need for the `org.scijava.event` package here.

PR #81, opened 2023-07-14
Narrow Collection returns to Set/List

### LINKED ISSUE #123: Narrow return interface of OpEnvironment descriptions methods
Right now, the `descriptions` methods of `OpEnvironment` return `Collection`. The rule of thumb for API design is for method parameters to use the broadest possible interface type, but for method return values to use the most specific possible interface type. So I suggest to change the `descriptions` methods to return `List` instead. And as part of this issue, quickly scan/grep over the entire codebase's API for similar issues and try to fix them opportunitistically.

gselzer:
Closes scijava/scijava#123

@ctrueden I did a quick grep `grep -r -E 'public\s+\w+\s+\w+\(\)\s+\{'` through the `scijava` folder of the incubator, and didn't find too much to narrow, but you can feel free to do your own checks if you'd like.

### DIFF
--- a/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java
+++ b/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java
@@ -30,7 +30,7 @@
 package org.scijava.ops.api;
 
 import java.lang.reflect.Type;
-import java.util.Collection;
+import java.util.Set;
 
 import org.scijava.types.Nil;
 
@@ -721,7 +721,7 @@ public interface OpEnvironment {
 	/**
 	 * Creates an {@link OpInfo} from an {@link Class}.
 	 *
-	 * @param opClass
+	 * @param opClass the {@link Class} to derive the Op from
 	 * @return an {@link OpInfo} which can make instances of {@code opClass}
 	 */
 	OpInfo opify(Class<?> opClass);
@@ -729,7 +729,7 @@ public interface OpEnvironment {
 	/**
 	 * Creates an {@link OpInfo} from an {@link Class} with the given priority.
 	 *
-	 * @param opClass
+	 * @param opClass the {@link Class} to derive the Op from
 	 * @param priority - the assigned priority of the Op.
 	 * @param names - the name(s) of the Op
 	 * @return an {@link OpInfo} which can make instances of {@code opClass}
@@ -746,9 +746,9 @@ public interface OpEnvironment {
 	/**
 	 * Creates some {@link OpInfo}s from {@code o}
 	 * @param o the {@link Object} to create {@link OpInfo}s from
-	 * @return a {@link Collection} of {@link OpInfo}s
+	 * @return a {@link Set} of {@link OpInfo}s
 	 */
-	Collection<OpInfo> infosFrom(Object o);
+	Set<OpInfo> infosFrom(Object o);
 
 	/**
 	 * Creates some {@link OpInfo}s from {@code o}
@@ -769,10 +769,10 @@ public interface OpEnvironment {
 	 * Note that this method does not inherently provide <b>any</b> guarantees
 	 * pertaining to thread-safety; this API is provided purely for convenience.
 	 * Any calls to {@link OpEnvironment#op(String, Nil, Nil[], Nil)} that require
-	 * a specific {@Hints} should <b>not</b> use this method, instead opting for
+	 * a specific {@link Hints} should <b>not</b> use this method, instead opting for
 	 * {@link OpEnvironment#op(String, Nil, Nil[], Nil, Hints)}.
 	 * 
-	 * @param hints
+	 * @param hints the {@link Hints} to be set as default
 	 */
 	void setDefaultHints(Hints hints);
 
@@ -797,16 +797,16 @@ public interface OpEnvironment {
 	 * Returns the descriptions for all Ops contained within this
 	 * {@link OpEnvironment}
 	 *
-	 * @return a {@link Collection} of descriptions
+	 * @return a {@link Set} of descriptions
 	 */
-	Collection<String> descriptions();
+	Set<String> descriptions();
 
 	/**
 	 * Returns the descriptions for all Ops identifiable by a given name within
 	 * this {@link OpEnvironment}
 	 *
-	 * @return a {@link Collection} of descriptions
+	 * @return a {@link Set} of descriptions
 	 */
-	Collection<String> descriptions(String name);
+	Set<String> descriptions(String name);
 
 }
--- a/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java
+++ b/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java
@@ -30,6 +30,7 @@
 package org.scijava.ops.api;
 
 import java.lang.reflect.Type;
+import java.util.List;
 import java.util.Set;
 
 import org.scijava.types.Nil;
@@ -59,13 +60,13 @@ import org.scijava.types.Nil;
 public interface OpEnvironment {
 
 	/** The available ops for this environment. */
-	Iterable<OpInfo> infos();
+	List<OpInfo> infos();
 	
-	Iterable<OpInfo> infos(String name);
+	List<OpInfo> infos(String name);
 
-	Iterable<OpInfo> infos(Hints hints);
+	List<OpInfo> infos(Hints hints);
 
-	Iterable<OpInfo> infos(String name, Hints hints);
+	List<OpInfo> infos(String name, Hints hints);
 
 	// TODO: Add interface method: OpInfo info(final String opName, final Nil<T> specialType, final Nil<?>[] inTypes, final Nil<?> outType);
 
@@ -799,7 +800,7 @@ public interface OpEnvironment {
 	 *
 	 * @return a {@link Set} of descriptions
 	 */
-	Set<String> descriptions();
+	List<String> descriptions();
 
 	/**
 	 * Returns the descriptions for all Ops identifiable by a given name within
@@ -807,6 +808,6 @@ public interface OpEnvironment {
 	 *
 	 * @return a {@link Set} of descriptions
 	 */
-	Set<String> descriptions(String name);
+	List<String> descriptions(String name);
 
 }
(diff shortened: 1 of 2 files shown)


scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java
```diff
@@ -746,9 +746,9 @@ default OpBuilder.Arity16 arity16(final String opName, final Hints hints) {
 	/**
 	 * Creates some {@link OpInfo}s from {@code o}
 	 * @param o the {@link Object} to create {@link OpInfo}s from
-	 * @return a {@link Collection} of {@link OpInfo}s
+	 * @return a {@link Set} of {@link OpInfo}s
 	 */
-	Collection<OpInfo> infosFrom(Object o);
+	Set<OpInfo> infosFrom(Object o);
```
ctrueden, line 752: Why `Set` rather than `List`? If we use `List` then we have an ordering, and we can return them ordered by priority, which is useful in some scenarios. Right now, `infosFrom` uses `Collectors.toSet`, you `descriptions` was changed to use a set also, but `opsFromObject` still uses `Collectors.toList`. I'd like to be very intentional here in each case about whether ordering would be a good thing or a bad thing, and type accordingly.

ctrueden (commented):
Before merging this, I'd like to better understand why we sometimes return `Set` and other times `List`. I lean toward `List` always unless the intended use case really benefits from `Set`—e.g. if there is no explicit ordering, or there can be unwanted duplicates otherwise, or if there will be a lot more individual inclusion tests (which are O(1) with sets) and fewer complete iterations (which are faster with list).

gselzer:
> Before merging this, I'd like to better understand why we sometimes return `Set` and other times `List`. I lean toward `List` always unless the intended use case really benefits from `Set`—e.g. if there is no explicit ordering, or there can be unwanted duplicates otherwise, or if there will be a lot more individual inclusion tests (which are O(1) with sets) and fewer complete iterations (which are faster with list).

That's a good point. Originally, I was concerned about unwanted duplicates i.e. what if some Op was discovered from two different `Discoverer`s, but we can avoid that using other means. Let me know what you think of 48bfe24

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
```diff
@@ -166,30 +194,32 @@ public static List<MatchingRoutine> getMatchingRoutines(
...
 	}
 
 	@Override
-	public Set<OpInfo> infos(String name, Hints hints) {
+	public List<OpInfo> infos(String name, Hints hints) {
 		return filterInfos(infos(name), hints);
 	}
 
-	private Set<OpInfo> filterInfos(Set<OpInfo> infos, Hints hints) {
+	private List<OpInfo> filterInfos(List<OpInfo> infos, Hints hints) {
```
ctrueden, line 222: Here's a place where the input `infos` could be a `Collection`, because all you do with them is call `.stream()`. So you should type the method argument directly on the interface that declares the `stream()` method.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
```diff
@@ -724,6 +754,13 @@ private synchronized void initIdDirectory() {
 				.forEach(info -> idDirectory.put(info.id(), info));
 	}
 
+	/**
+	 * Function for populating opDirectory with new ArrayList.
+	 * Declared only once, right here
+	 */
+	private final Function<String, List<OpInfo>> nameListCreator = //
+		(name) -> new ArrayList<>();
```
ctrueden: I think the parentheses around the `name` parameter are unneeded.

ctrueden: Also: is it actually needed to have this function, for performance? Does Ops behave in a worse way if you just inline this lambda below where the `nameListCreator` is passed to `computeIfAbsent`? I'm skeptical that the JIT doesn't just deal with it...

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
```diff
@@ -348,10 +378,10 @@ public Set<String> descriptions(String name) {
 	 * @return a {@link Set} of {@link String}s, one describing each
 	 *         {@link OpInfo} in {@code infos}.
 	 */
-	private Set<String> descriptions(Iterable<OpInfo> infos) {
+	private List<String> descriptions(Iterable<OpInfo> infos) {
 		return StreamSupport.stream(infos.spliterator(), true) //
 				.map(OpInfo::toString) //
-				.collect(Collectors.toSet());
+				.collect(Collectors.toList());
```
ctrueden: I saw that you use `Collectors.toUnmodifiableList()` below... would it make sense to do so also here?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
```diff
@@ -201,7 +231,7 @@ private Set<OpInfo> filterInfos(Set<OpInfo> infos, Hints hints) {
 			// filter out unadaptable ops
 			.filter(info -> !simplifying || !info.declaredHints().contains(
 				Simplification.FORBIDDEN)) //
```
ctrueden, line 233: AGh, maybe it's time we increase the max code line length from 80 to 120...

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
```diff
@@ -42,11 +53,28 @@
...
+import org.scijava.ops.api.OpDependencyMember;
+import org.scijava.ops.api.OpEnvironment;
+import org.scijava.ops.api.OpHistory;
+import org.scijava.ops.api.OpInfo;
+import org.scijava.ops.api.OpInfoGenerator;
+import org.scijava.ops.api.OpInstance;
+import org.scijava.ops.api.OpMetadata;
+import org.scijava.ops.api.OpRef;
+import org.scijava.ops.api.OpWrapper;
+import org.scijava.ops.api.RichOp;
```
ctrueden, line 69: Hello imports! :wave:

ctrueden (changes requested):
Nice! I made a couple of small suggestions. Let me know what you think!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
ctrueden: And/or, you could make the line shorter by doing a static import on the `FORBIDDEN` constant...

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
gselzer: That's true, it could, but it's also a private method, and it makes things simpler within the public functions, so I'd say we leave it as is?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
gselzer: Probably not - inlined in 63a1210a

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
gselzer: Yeah, there are unfortunately a lot of wildcard imports (`git grep import.*\*`) - maybe we should do a pass through at some point to get rid of the rest of them.

gselzer: Made https://github.com/scijava/scijava/issues/126 to track it.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
gselzer: > And/or, you could make the line shorter by doing a static import on the `FORBIDDEN` constant...

Don't think that's possible - we use both `Simplification.FORBIDDEN` and `Adaptation.FORBIDDEN` - we'd have a clash.

> AGh, maybe it's time we increase the max code line length from 80 to 120...

I don't feel too strongly - I do think there's something to be said about shorter line lengths - maybe some of our API is too verbosely named? **But** now is the time to change it, maybe we should file an issue about a SciJava 3 code style?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
gselzer: Probably - I like the idea of preventing users from writing to these returns and thinking they're adding a new Op.

Changed in 9c6bd502

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
```diff
@@ -773,7 +766,7 @@ private synchronized void initIdDirectory() {
 			return;
 		}
 		for (String opName : opInfo.names()) {
-			opDirectory.computeIfAbsent(opName, nameListCreator).add(opInfo);
+			opDirectory.computeIfAbsent(opName, name -> new ArrayList<>()).add(opInfo);
```
ctrueden, line 769: Oh wait no... do we need a field for generic typing reasons?! The lambda won't be inspectable by the ops engine, will it?

ctrueden (commented):
I was going to merge this now, but then I realized I maybe steered us wrong with that "inline the lambda" suggestion. What do you think?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/DefaultOpEnvironment.java
ctrueden: Or is this not an op so it doesn't matter? Sorry it's late and I'm on mobile, probably I should just scrutinize more carefully myself on a real computer. Anyway, if the lambda generics aren't actually a problem here then I fully support merging this immediately.

gselzer: This isn't an Op, no worries :sweat_smile:

All this function does is adds a new `List` to track Ops named `name` if one doesn't exist already

ctrueden (approved):
LGTM!

PR #82, opened 2023-07-25
Construct static Logging utitity class

### LINKED ISSUE #13: Decide how to support logging errors and warnings
Many ops in the old imagej-ops framework, such as [this one](https://github.com/imagej/imagej-ops/blob/1820948f097f098878cf5d5745d39c63cf2e6049/src/main/java/net/imagej/ops/threshold/minError/ComputeMinErrorThreshold.java#L59) or [this one](https://github.com/imagej/imagej-ops/blob/1820948f097f098878cf5d5745d39c63cf2e6049/src/main/java/net/imagej/ops/threshold/rosin/ComputeRosinThreshold.java#L59) attempt to log errors or warnings. Sometimes this is in place of throwing some `RuntimeException` (as in the case of the former) or other times it is just saying something about the data (as in the case of the latter). We need to find a consistent way to deal with the situations in which we would like to pass on messages/errors.

Regardless of the option, no messages should be allowed as secondary outputs.

**Potential solutions:**
* Use Scijava's `Logger`
* Create dedicated error subclasses (although this will only solve the issues with errors, not with the logging)
* Not allow any sort of logging

### LINKED ISSUE #64: Log Warnings for Ops without proper javadocs
To promote thoroughly documented Ops, we should log a warning whenever our Javadoc parser determines that there is not enough information to *at least* name all of the parameters (there is also the question of descriptions, however that is likely too draconian a request). The logging should be relatively straightforward, and occur in `ParameterStruct`s, however this issue will have to wait on #13.

### LINKED ISSUE #106: Use SciJava Log2 Logger within SciJava Discoverer implementations
Each `Discoverer` is responsible for reporting an arbitrary number of  implementations for each call to `Discoverer.discover()`.

Sometimes, these `Discoverer`s might find that one of the implementations is incorrect in some way. Instead of throwing an error, I think the principle of least surprise suggests that they should notify the user and move on, returning the correct discoveries instead of none at all. SciJava Log2's `Logger` class would be perfect for this.

Unfortunately, though, we don't provide the API for `Discoverer`s to access a `Logger`, so as @ctrueden noticed `Discoverers` either use `Throwable.printStackTrace` or just use `System.out.println` or the equivalent.

I don't have a great idea of how to solve this. Off the top of my head we could:
* Add a `Logger` to the `discover` function. I dislike this, I don't like metaparameters leaking into signatures.
* Make a `LoggingDiscoverer` interface with getters/setters for a `Logger`.
* Make a static method somewhere that each `Discoverer` could call to log errors. Would different `Discoverer`s ever want to log to different outputs? If so, this probably is a bad idea.

gselzer:
This prevents us from having to pass a Logger around everywhere!

In addition, I noticed that the `Logger` and `OpMonitor` utility classes were still hanging around in the engine, so I removed them!

Closes scijava/scijava#106, scijava/scijava#13 and maybe other issues.


gselzer:
With e000bd72, we can also close scijava/scijava#64

gselzer:
Over in scijava/scijava#13, @ctrueden said:

> here are valid reasons for wanting to log, for example, only a single op at debug, but everything else at info or warn, because you really just want to debug that one op implementation. Being able to do this without changing/rebuilding any code can be really important in some deployment scenarios. 

With this PR, you'd do something like the following:
```java
var theExistingLogger = Logging.getLogger();
Logging.setLogger(myVeryVerboseLogger);
ops.op("some op I want to debug").arityX().inputs(...).apply();
Logging.setLogger(theExistingLogger);
```
What would you prefer? Something like:
```java
var op = ops.op("some op I want to debug").arityX().inputs(...).function();
Logging.setLogger(op, myVeryVerboseLogger);
op.apply(...)
```
I don't feel too strongly one way or the other - honestly, it wouldn't be too terribly hard to  add the other, if that's what you think people will use.

> There are also valid reasons for wanting to have a different logger used per OpEnvironment, in case you have some complicated server-side JVM running multiple separate contexts.

This seems like something that the `AbstractRichOp` could be responsible for doing. Do you really foresee someone doing this? I'm happy to punt on this until it's needed.

> From an architecture standpoint, I can't say that I love having a static Logging utility class—I think it limits the utility of logging—and we might be better off just going with a "standard best practice" logging approach that doesn't try to be too clever about reducing boilerplate. E.g. just suck it up and put the SLF4J Logger declaration at the top of each class that wants to do logging, I dunno...

Isn't [this](https://www.slf4j.org/manual.html#typical_usage) basically what we're doing now?

hinerm:
Discussion in https://github.com/scijava/scijava/issues/13 looks unresolved. I'm going to move this to a draft PR and remove the reviewers for the moment until we have a chance to discuss further.

PR #83, opened 2023-07-26
Throw ValidityException instead of saving it

gselzer:
This PR simplifies the `OpInfo` API around `ValidityException`s - instead of saving the `Exception`, the constructor now throws the `Exception`. Then, during `OpInfo` construction, we now catch these `Exceptions` and continue on.

We still have to figure out how to log the `Exceptions`, so that developers can determine how to make their ops valid - I'm tempted to just wait until #82 is merged, because logging would then be quite easy.

Question for the reviewer: Are failures getting tested? Should we add additional tests? What might those tests look like?

### DIFF
--- a/scijava/scijava-common3/src/main/java/org/scijava/common3/validity/ValidityException.java
+++ b/scijava/scijava-common3/src/main/java/org/scijava/common3/validity/ValidityException.java
@@ -10,7 +10,7 @@ import java.util.List;
  * @author Curtis Rueden
  * @author Gabriel Selzer
  */
-public final class ValidityException extends Exception {
+public final class ValidityException extends RuntimeException {
 
 	private final List<ValidityProblem> problems;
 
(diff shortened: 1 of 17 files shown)


hinerm (approved):

hinerm:
>Are failures getting tested? Should we add additional tests? What might those tests look like?

The failures are not getting tested (per uses of [ValidityException.problems()](https://github.com/scijava/incubator/blob/28755f92eed318485df89e0b165bcc162ccfb624/scijava/scijava-common3/src/main/java/org/scijava/common3/validity/ValidityException.java#L26)) except in the recent [OpMethodDependencyPositionTest](https://github.com/scijava/incubator/blob/28755f92eed318485df89e0b165bcc162ccfb624/scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/OpMethodDependencyPositionTest.java#L109).

I think the tests should be looked at with https://github.com/scijava/scijava/issues/105, because `ValidityProblem`s still need some API work.

PR #84, opened 2023-08-02
Refactor TypeExtractor

gselzer:
This PR refactors the `TypeExtractor` interface, providing much more power in the ability to determine the type of an `Object`.

The main use case driving this refactor is the ability to control the reification of types like `net.imagej.Dataset` - types that have no type variables, who we'd like to know the type variables for. This is vital for using ImageJ Ops2 in SciJava scripts, because we need to be able to reify `Dataset`s into a `Img<T>`.

Still need to fix the ImageJ Ops2 tests - probably a bug in one of the converted `TypeExtractor`s.


imagej/imagej-ops2/src/main/java/net/imagej/ops2/filter/gauss/Gaussians.java
```diff
@@ -71,11 +70,11 @@ public class Gaussians {
 	 *           type='org.scijava.function.Computers$Arity4'
 	 */
 	@SuppressWarnings({ "unchecked", "rawtypes" })
-	public static <T extends NumericType<T> & NativeType<T>> void defaultGaussRAI(
-		final RandomAccessibleInterval<T> input, //
+	public static <S extends NumericType<S>, T extends NumericType<T>> void defaultGaussRAI(
+		final RandomAccessibleInterval<S> input, //
```
ctrueden: How about using `EI` for element type of input image, and something else consistent... `EB`?... for element type of `OutOfBoundsFactory`? We use `EI` for element type of input images and `EO` for element type of output images in ImageJ Ops v1 in a few places, and I'd like to see more consistency here if possible, to reduce possible confusion.

imagej/imagej-ops2/src/main/java/net/imagej/ops2/types/ImgFactoryTypeExtractor.java
```diff
@@ -48,25 +49,21 @@
...
+public class ImgFactoryTypeExtractor implements TypeExtractor {
 
-	@Override
-	public Type reify(final TypeReifier t, final ImgFactory<?> o, final int n) {
-		if (n != 0)
-			throw new IndexOutOfBoundsException();
-
-		return t.reify(o.type());
+	@Override public double getPriority() {
+		return Priority.NORMAL;
```
ctrueden: Isn't `NORMAL` already the default? If not, shouldn't it be?

imagej/imagej-ops2/src/main/java/net/imagej/ops2/types/ImgFactoryTypeExtractor.java
```diff
@@ -48,25 +49,21 @@
...
 
-	@Override
-	public double priority() {
-		return Priority.LOW;
+	@Override public Type reify(TypeReifier r, Object object) {
+		if (!(object instanceof ImgFactory))
+			throw new IllegalArgumentException(this + " cannot reify " + object);
+		ImgFactory<?> imgFactory = (ImgFactory<?>) object;
+		Type componentType = r.reify(imgFactory.type());
+		return TypeTools.raiseParametersToClass(object.getClass(), ImgFactory.class, new Type[] {componentType});
```
ctrueden: In general, I love the generalization of the `reify` method. But I dislike this boilerplate `TypeTools.raiseParametersToClass` thing that is now needed in the majority of `TypeExtractor` implementations. Surely we can make the code more DRY somehow?

scijava/scijava-types/src/main/java/org/scijava/types/TypeTools.java
```diff
@@ -302,6 +304,20 @@ else if (paramComponent instanceof TypeVariable) {
 		return true;
 	}
 
+	public static Type raiseParametersToClass(Class<?> cls, final Class<?> superCls, final Type[] superClsTypeVars) {
```
ctrueden: The `superClsTypeVars` could be a `Type...` instead, to make calling code a little more readable.

I would also rename this method to something less bulky, especially since it is called by nearly every `TypeExtractor` implementations as its final step.

scijava/scijava-types/src/main/java/org/scijava/types/extractors/ParameterizedTypeExtractor.java
```diff
@@ -0,0 +1,198 @@
...
+				// of the generic type hierarchy, which we don't care about here.
+				for (final TypeVariable<?> typeVar : typeVars) {
+					if (vars.containsKey(typeVar)) {
+						resolved.putIfAbsent(typeVar, vars.get(typeVar));
+					}
+				}
+			}
+		}
+
+		// fill in any remaining unresolved type parameters with wildcards
```
ctrueden: This comment style differs from earlier comments. Stick with one style.

ctrueden (commented):
Nice progress! I made some comments.

imagej/imagej-ops2/src/main/java/net/imagej/ops2/types/ImgFactoryTypeExtractor.java
gselzer: Yeah, it should be. Fixed in 877d83f

imagej/imagej-ops2/src/main/java/net/imagej/ops2/types/ImgFactoryTypeExtractor.java
gselzer: Made an abstract superclass in 4e897c7c

imagej/imagej-ops2/src/main/java/net/imagej/ops2/filter/gauss/Gaussians.java
gselzer: A quick grep search shows no instances of `EI`/`EB`/`EO` in the incubator, so I'd prefer to not introduce it.

I used `S` and `T` in this class because that's what the underlying `SeparableSymmetricConvolution` uses, but we could change it to `I`/`O`!

gselzer: Renamed in 035e1c1

scijava/scijava-types/src/main/java/org/scijava/types/TypeTools.java
gselzer: Renamed in 035e1c1

scijava/scijava-types/src/main/java/org/scijava/types/extractors/ParameterizedTypeExtractor.java
gselzer: I'm not fully sure what you meant, but I tried to fix it in 035e1c1

ctrueden: I meant:
```
// fill in any remaining unresolved type parameters with wildcards
```
vs.
```
// Fill in any remaining unresolved type parameters with wildcards.
```

So, whether to write in complete sentences, or only small case, basically.

For many years I was a small-case no-period guy, but I kind of came around to the idea that reading comments should be like reading anything else written by someone: so make it complete sentences in most situations.

But that's only my personal preference. The main thing is just to be consistent—and I noticed that we had some of the former and others of the latter.

It's a nitpick for sure, though!~

hinerm (approved):

PR #88, opened 2023-08-22
Replace ValidityProblem usage with InvalidOpExceptions

### LINKED ISSUE #105: Make use of ValidityProblems during Op matching
With the merge of https://github.com/scijava/incubator/commit/2adac432f5efa9c85a0932548810121b30cf3306 we now have `@Optional` parameters. This comes with the caveat that parameters should only be declared in **one** location, i.e. *either* an op interface or method implementation, but not both.

The [optional-op-param-validation](https://github.com/scijava/incubator/tree/scijava/scijava-ops-engine/optional-op-param-validation) adds tests creating this ambiguous state.

In ops, [this state is currently detected](https://github.com/scijava/incubator/blob/2adac432f5efa9c85a0932548810121b30cf3306/scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/struct/LazilyGeneratedMethodParameterData.java#L235-L249) and a [ValidityProblem](https://github.com/scijava/incubator/blob/main/scijava/scijava-common3/src/main/java/org/scijava/common3/validity/ValidityProblem.java) is created to record the issue.

However, the actual *call* into detecting parameter optionality is just passing [an anonymous list](https://github.com/scijava/incubator/blob/2adac432f5efa9c85a0932548810121b30cf3306/scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/struct/LazilyGeneratedMethodParameterData.java#L116-L117) that is immediately discarded. This results in op requests failing with no information actually provided.

We should, at the least, be actually propagating `ValidityProblems` and including them in any "No candidates found" report at the end of a failed matching session.

^ this may not be necessary any more since https://github.com/scijava/incubator/pull/83

But I suggest the following:

- [ ] Clean up dead usages of `problems` lists [like this one](https://github.com/scijava/incubator/blob/28755f92eed318485df89e0b165bcc162ccfb624/scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/reduce/ReducedOpInfo.java#L42-L44).
- [ ] Turn `ValidityProblem` to an enum and use that instead of message [string comparison](https://github.com/scijava/incubator/blob/28755f92eed318485df89e0b165bcc162ccfb624/scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/OpMethodDependencyPositionTest.java#L109)
- [ ] Add a test for each other `ValidityProblem` type, modeled after the `OpMethodDependencyPositionTest`

gselzer:
This PR makes Ops throw `InvalidOpException`s instead of `ValidityException`s - this creates a simpler structure for conveying issues with Op configuration/declaration to Op developers. It also provides a suite of tests to ensure invalid Ops are being 

TODO: @hinerm - you said [here](https://github.com/scijava/scijava/issues/105#issuecomment-1686751032) that you do not want string comparison - can you elaborate on why? Would you prefer that we subclass `InvalidOpException` for each of the reasons of failure? There would be a lot of subclasses with pretty detailed names, if you'd prefer that...


gselzer:
This will close scijava/scijava#105

hinerm:
>you said https://github.com/scijava/scijava/issues/105#issuecomment-1686751032 that you do not want string comparison - can you elaborate on why? Would you prefer that we subclass InvalidOpException for each of the reasons of failure?

@gselzer so the concern I had was using raw string values as a substitute for testing for the type of failure. If the error's message changes for any reason the test will fail. What you're testing is that the error message is consistent, but what you actually want to test is that the type of error is consistent.

Subclassing and checking the subclass in the test would probably be the most Java-correct thing to do. Using constants, like in [BaseOpHints](https://github.com/scijava/incubator/blob/6ac889150f9911bf6fd02a2d87de8af22dd50750/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/features/BaseOpHints.java#L11-L42), instead of raw strings in the test expected values would be an acceptable improvement.

gselzer:
@hinerm I made a bunch of subclasses, let me know what you think of this.

PR #89, opened 2023-08-22
Simplify hints by moving implementation upstream

### LINKED ISSUE #82: Finalize naming of hints
I find the naming confusing: we have a `DefaultHints` *class* and `getDefaultHints`/`setDefaultHints` *methods*, and the methods do not operate on `DefaultHints` objects (necessarily), but rather the interface. It's a name clash conceptually between our usage of `Default` as a prefix for the default implementation for interfaces, and "default" as a technical term meaning "the settings which, all other things being equal and unspecified, we use to operate."

It's been a long time since the hints API has taken up any of my brain space, so I may have reverted back to naivety here, but in a perfect world I think `Hints` should be a single class with no interface at all. And anyone who wants to modify how it works (which should be vanishingly rare), would just subclass it.

gselzer:
This PR moves the `DefaultHints` implementation upstream, as because @ctrueden points out in scijava/scijava#82, the existence of a `DefaultHints` implementation is confusing, and because we haven't yet found a use case where someone would want multiple implementations. *Of course, this change doesn't prevent subclassing - maybe it will in fact make it easier*


hinerm (approved):
I even like the improved lambdas 😂

PR #64, opened 2022-10-21
Scijava/scijava ops engine/skip map history

hinerm:
Closes https://github.com/scijava/scijava/issues/99

See also [#100](https://github.com/scijava/scijava/issues/100) and [#101](https://github.com/scijava/scijava/issues/101)

PR #90, opened 2023-08-28
Minimize and proof SciJava Ops API & SPI

### LINKED ISSUE #73: Consider separating SciJava Ops API into two (or more) modules
The [migrate-features-to-engine branch](https://github.com/scijava/incubator/tree/scijava/scijava-ops-api/migrate-features-to-engine) introduces a `features` package into SciJava Ops API, to house "more theoretical" implementation details (i.e. they are only tied to an implementation in spirit). Since the classes living here aren't _quite_ API, we might eventually want another module to house this.

### LINKED ISSUE #81: Finalize naming of ops-related core classes
`OpInfo`, `OpInstance`, `InfoChain`, `RichOp` – as well as what objects we are going to name "op". Is it only the raw objects that are instances of functional interfaces? Or can we name instances of any of the above classes "op"? In the docs, which types of things do we refer to as "op"?

gselzer:
This PR reduces the footprint of the SciJava Ops API, moving implementation-specific classes to other components (mostly to SciJava Ops Engine) - this reduces the burden of maintaining our exposed API.

To enable testing API functionality, I've created a module `scijava-ops-test`, which allows runtime access to `DefaultOpEnvironment`, but only a compile-time dependency on `scijava-ops-api`. This module will prove that the API is powerful enough to access all functionality housed in the incubator.

To ensure that the SPI is suitable, I've tried to remove the engine dependency from ImageJ Ops2 - unfortunately, it still requires an API dependency, although maybe we can fix this later!

TODO:
* [x] Clean up `scijava-ops-test` - do we really need a separate module? We might want to consider making this an integration test package in the engine...
* [x] Determine why we needed to `open` the `net.imagej.ops2` module to get the tests to pass... is this something that rebasing over #79 might fix?
* [x] `net.imagej.ops2.OpRegressionTest` reports finding 2 fewer Ops - did we actually remove some Ops?
* [x] Can we remove the `org.scijava.ops.api` dependency from imagej ops2? It looks like only the `map` Ops use it to avoid recording the history of dependencies - relevant topics may include #33, #64 - **EDIT:** No, we can't, right now, without removing functionality. Skipping this...
* [x] Rename `DefaultMatchingErrorTest` to `OpMatchingExceptionTest`

Closes scijava/scijava#73


gselzer:
> Determine why we needed to `open` the `net.imagej.ops2` module to get the tests to pass... is this something that rebasing over

To add some context, here's what happens - in the tests, we use therapi to discover `OpInfo`s from imagej ops2, and the therapi javadoc parser calls loader.getResourcesAsStream("/" + resourceName), where `resourceName` is a `__Javadoc.json` file and `loader` is a class within the imagej.ops2 module. For some reason, this does a `Module.opens` check, ensuring that `net.imagej.ops2` is open *to the unnamed module* - I have no idea why we'd need this - would be worth checking whether this same check was made before...

gselzer:
TODO: Make changes described in [this comment](https://github.com/scijava/scijava/issues/81#issuecomment-1735922273)

PR #91, opened 2023-09-06
Add tests for ConvertTypes

### LINKED ISSUE #112: Add unit tests for ConvertTypes
I didn't see any explicit calls to the converters in [ConvertTypes](https://github.com/scijava/incubator/blob/ad1535415a4dd36eb5837ccde9e6e7bca749e42f/imagej/imagej-ops2/src/main/java/net/imagej/ops2/convert/ConvertTypes.java#L61)... are they being tested indirectly?

If not, they seem fairly fundamental and should be tested.

gselzer:
This PR adds in the autogenerated tests made for ImageJ Ops by @awalter17, massaged to work on ImageJ Ops2. This PR also massages/eliminates some `create` Ops to this end, to try and coalesce all Ops under the singular `create` namespace.

Points of discussion for the reviewers:
* These tests call the Ops as functions, but the Ops are written as `Computer.Arity1`s (this is slick for whole-image conversion). Unfortunately, for single-type conversion, though, this is tricky to match - you can't write `ops.unary("convert.float64").input(myByte).apply()`, because adaptation will fail*. As a result I've added `.outType` calls, however I think it looks unnecessarily clunky. The only other alternative I could think of, though, would be to have both a `Computer` and a `Function` implementation for all `convert` Ops, which defeats the purpose of adaptation :frowning_face: 
* Note that the first commit will fail the build because I did not solve the matching problems in the test until the second commit - we should squash the two commits before merging.

\* fails because the standard `Function` --> `Computers.Arity1` adaptors look for either a `Function<ByteType, Object>` or a `Producer<Object>`, which in the former case fails because a matched dependency makes the adaptation fail, and in the latter fails because there are multiple matches.

Closes scijava/scijava#112


hinerm:
>As a result I've added .outType calls, however I think it looks unnecessarily clunky

@gselzer agree that this is unfortunate but.. it works... it's too bad you can't just do something like:

```
ops.unary("convert").input(myByte).outType(Float64.class).apply()
```

>The only other alternative I could think of, though, would be to have both a Computer and a Function implementation for all convert Ops, which defeats the purpose of adaptation

yeah let's not go down this road right now

gselzer:
> > As a result I've added .outType calls, however I think it looks unnecessarily clunky
> 
> @gselzer agree that this is unfortunate but.. it works... it's too bad you can't just do something like:
> 
> ```
> ops.unary("convert").input(myByte).outType(Float64.class).apply()
> ```

Yup, I use the `Class` version of `outType`.

> > The only other alternative I could think of, though, would be to have both a Computer and a Function implementation for all convert Ops, which defeats the purpose of adaptation
> 
> yeah let's not go down this road right now

Cool. Merging!

PR #92, opened 2023-10-18
Make InfoTreeGenerator Prioritized

### LINKED ISSUE #87: Consider whether InfoChainGenerator should be extensible
And if so: whether there is ever a possibility of multiple implementations handling the same scenario. As [written currently](https://github.com/scijava/incubator/pull/46#discussion_r763345469), an exception would be thrown in that case.

gselzer:
@ctrueden noted [here](https://github.com/scijava/incubator/pull/46#discussion_r763345469) that current logic could throw errors if there are multiple InfoTreeGenerators that match a particular Op signature, and outlined two choices:
1. Hide `InfoTreeGenerator`, so people can't implement it, and hardcode the priority.
2. Implement a priority system, and when multiple `InfoTreeGenerator`s could reify the signature, choose the highest priority.

This PR chooses option 2 as it is the less-API-breaking change, and because a priority-based system makes more sense than encoding an ordering by e.g. position in a `List`. This PR *doesn't* hide `InfoTreeGenerator`, but it also does not prevent us doing so in the future, and I'm not particularly against hiding it until we decide we need it.

Hiding `InfoTreeGenerator` would be as simple as moving it to a non-exported package. I still believe we'd want to service-load the implementations, as otherwise we'd probably introduce sub-package dependencies, failing the build, but that should still work if we move the package of the `InfoTreeGenerator`.

Closes scijava/scijava#87

### DIFF
--- a/scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/InfoTreeGenerator.java
+++ b/scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/InfoTreeGenerator.java
@@ -36,8 +36,10 @@ import java.util.stream.Collectors;
 
 import org.scijava.ops.api.InfoTree;
 import org.scijava.ops.api.OpInfo;
+import org.scijava.priority.Prioritized;
+import org.scijava.priority.Priority;
 
-public interface InfoTreeGenerator {
+public interface InfoTreeGenerator extends Prioritized<InfoTreeGenerator> {
 
 
 	/**
@@ -68,13 +70,11 @@ public interface InfoTreeGenerator {
 	{
 		List<InfoTreeGenerator> suitableGenerators = generators.stream() //
 			.filter(g -> g.canGenerate(signature)) //
+			.sorted() //
 			.collect(Collectors.toList());
-		if (suitableGenerators.size() == 0) throw new IllegalArgumentException(
+		if (suitableGenerators.isEmpty()) throw new IllegalArgumentException(
 			"No InfoTreeGenerator in given collection " + generators +
 				" is able to reify signature " + signature);
-		if (suitableGenerators.size() > 1) throw new IllegalArgumentException(
-			"Signature " + signature +
-				" is able to be reified by multiple generators: " + suitableGenerators);
 		return suitableGenerators.get(0);
 	}
 
@@ -90,8 +90,8 @@ public interface InfoTreeGenerator {
 	 * Finds the subsignature in {@link String} {@code signature}. The
 	 * subsignature is assumed to start at index {@code start}.
 	 * 
-	 * @param signature
-	 * @param start
+	 * @param signature the signature containing a subsignature
+	 * @param start the index where the subsignature starts
 	 * @return a signature contained withing {@code signature}
 	 */
 	static String subSignatureFrom(String signature, int start) {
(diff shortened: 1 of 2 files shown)


scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/InfoTreeGenerator.java
```diff
@@ -124,4 +123,8 @@ else if (ch == InfoTree.DEP_END_DELIM) {
 	 */
 	boolean canGenerate(String signature);
 
+	@Override
+	default double getPriority() {
+		return Priority.NORMAL;
```
ctrueden, line 128: @gselzer Make it makes sense for the `Prioritized` interface itself to have this default?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/InfoTreeGenerator.java
```diff
@@ -68,13 +70,11 @@ static InfoTreeGenerator findSuitableGenerator(String signature,
...
+			.sorted() //
 			.collect(Collectors.toList());
-		if (suitableGenerators.size() == 0) throw new IllegalArgumentException(
+		if (suitableGenerators.isEmpty()) throw new IllegalArgumentException(
 			"No InfoTreeGenerator in given collection " + generators +
 				" is able to reify signature " + signature);
-		if (suitableGenerators.size() > 1) throw new IllegalArgumentException(
-			"Signature " + signature +
-				" is able to be reified by multiple generators: " + suitableGenerators);
 		return suitableGenerators.get(0);
```
ctrueden, line 78: What if the top two `InfoTreeGenerator` instances have the same priority? Should we fail fast? (I don't have an opinion, just pointing out that right now, on ties, one of them arbitrarily wins.)

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/InfoTreeGeneratorTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.scijava.ops.api.InfoTree;
+import org.scijava.ops.api.OpInfo;
+import org.scijava.ops.engine.matcher.impl.OpFieldInfo;
+import org.scijava.ops.spi.OpCollection;
+import org.scijava.ops.spi.OpField;
+import org.scijava.priority.Priority;
+
+public class InfoTreeGeneratorTest extends AbstractTestEnvironment implements
```
ctrueden, line 17: :test_tube: :test_tube: :test_tube:

ctrueden (commented):
:+1: OK!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/InfoTreeGenerator.java
gselzer: I kinda like the fact that it doesn't - it makes you think about what the priority should be amongst the existing implementations. And there's the fact that this was the smallest necessary change to avoid our bug.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/InfoTreeGenerator.java
gselzer: I'm going to say YAGNI for now

PR #94, opened 2023-10-26
Describe levels of Op equality

### LINKED ISSUE #36: Document equality/equivalence semantics
Historically, it has been a requirement of ops that _two ops with the same name accepting the same inputs should produce equivalent output_, where "equivalent output" means equality of content. This has nice consequences, including facilitating reproducibility and caching. However, there are questions:

It is too draconian with respect to "create" ops that implement `Source`, since two ops of the same name may return different sorts of objects that are not comparable with one another, given the same inputs (that is to say: no inputs).

If you consider the output type as part of the "`OpRef` uniqueness signature" then there is no conflict, because the output types differ. We could have e.g. a `LongType create()` with `names="create, create.integerType, create.longType"` and `DoubleType create()` with `names="create, create.floatingType, create.doubleType` (at higher priority, so that a request for a "create" op of type `Source<T extends RealType<T>>` prefers the `DoubleType` creator).

However, "equality of content" is subtle and complex. Do a `LongType` and a `DoubleType` have equal content if their `getRealDouble()` methods return the same value? Do a `CellImg` and an `ArrayImg` have equal content if they are the same type with same values and same dimensionality? Intuitively that was the idea, although it is important to note that the will not compare as equal via `equals`—those classes do not override `equals`. This invites the possibility of additional new ops: `equals` or perhaps `equalContent` that tests two objects for this sort of content equality; and `compatible` or perhaps `isContentComparable` that discerns whether it even makes sense to consider this question across two types.

We already have `copy` computer ops that require you to specify the preallocated output object into which the content of the input will be copied. It would make sense to formally state that doing `equalContent(in, out)` after calling `copy(in, out)` should always return `true`; similarly, if `isContentComparable(in, out)` returns `false` then calling `copy(in, out)` should not find any ops to perform the work, since it makes no sense.

My feeling is that there are many subtleties and edge cases surrounding these behaviors, so: A) these design decisions are crucial, and we need to write them down as formally as possible; and B) where feasible, we need to programmatically validate them in unit tests etc.

gselzer:
Closes scijava/scijava#36

### DIFF
--- a/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Ops.java
+++ b/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Ops.java
@@ -26,6 +26,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  * #L%
  */
+
 package org.scijava.ops.api;
 
 /**
@@ -38,7 +39,9 @@ public final class Ops {
 	private Ops() {}
 
 	/**
-	 * Convenience function for determining whether {@code op} is a {@link RichOp}.
+	 * Convenience function for determining whether {@code op} is a
+	 * {@link RichOp}.
+	 * 
 	 * @param op the Op
 	 * @return true iff {@code op} is a {@link RichOp}
 	 */
@@ -48,20 +51,28 @@ public final class Ops {
 
 	/**
 	 * Convenience function for getting the {@link RichOp} of {@code op}
+	 * 
 	 * @param op the Op
 	 * @return the {@link RichOp} wrapping {@code op}
 	 * @param <T> the type of {@code op}
+	 * @throws IllegalArgumentException when a {@link RichOp} cannot be obtained
+	 *           for {@code op}
 	 */
 	@SuppressWarnings("unchecked")
 	public static <T> RichOp<T> rich(T op) {
+		if (!isRich(op)) {
+			throw new IllegalArgumentException(op + " is not a RichOp!");
+		}
 		return (RichOp<T>) op;
 	}
 
 	/**
 	 * Convenience function for getting the {@link OpInfo} of {@code op}
+	 * 
 	 * @param op the Op
 	 * @return the {@link OpInfo} that generated {@code op}
 	 * @param <T> the type of {@code op}
+	 * @throws IllegalArgumentException if {@code op} is not an Op
 	 */
 	public static <T> OpInfo info(T op) {
 		return rich(op).instance().infoTree().info();
@@ -73,6 +84,7 @@ public final class Ops {
 	 * @param op the Op
 	 * @param record true iff {@code op} should record its executions
 	 * @param <T> the type of the Op
+	 * @throws IllegalArgumentException if {@code op} is not an Op
 	 */
 	public static <T> void recordExecutions(T op, boolean record) {
 		rich(op).recordExecutions(record);
--- a/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Ops.java
+++ b/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Ops.java
@@ -58,6 +58,16 @@ public final class Ops {
 	}
 
 	/**
+	 * Convenience function for getting the {@link OpInfo} of {@code op}
+	 * @param op the Op
+	 * @return the {@link OpInfo} that generated {@code op}
+	 * @param <T> the type of {@code op}
+	 */
+	public static <T> OpInfo info(T op) {
+		return rich(op).instance().infoTree().info();
+	}
+
+	/**
 	 * Convenience function for accessing {@link RichOp#recordExecutions(boolean)}
 	 *
 	 * @param op the Op
--- a/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
+++ b/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
@@ -14,7 +14,7 @@
  * To support the Op matching paradigm, we establish three types of equivalence:
  * <ol>
  *   <li><b>Form Equivalence</b></li> Drawn from <a href="https://en.wikipedia.org/wiki/Theory_of_forms">Plato's Theory of Forms</a>, Form Equivalence implies that two objects (which could be Ops, or Op parameters) theoretically draw from the same shared idea (such as an addition operation, or an image, etc.)
- *   <li><b>Op Equivalence</b></li> Algorithmic Equivalence means that two Ops:
+ *   <li><b>Algorithmic Equivalence</b></li> Algorithmic Equivalence means that two Ops:
  *   <ol>
  *     <li>Are Form Equivalent</li>
  *     <li>Have the same number of inputs and the same number of outputs</li>
@@ -25,8 +25,8 @@
  * </ol>
  * Within the Ops API, each type of equivalence is utilized in the following ways:
  * <ol>
- *   <li><b>Form Equivalence</b></li> If two Ops are "Form Equivalent", they are defined under the same name.
- *   <li><b>Op Equivalence</b></li> If two Ops are "Op Equivalent", then they are considered to implement the same computational algorithm.
+ *   <li><b>Form Equivalence</b></li> If two Ops are "Form Equivalent", they are defined under the same name, i.e. {@code op1.name().equals(op2.name())}
+ *   <li><b>Algorithmic Equivalence</b></li> If two Ops are "Algorithm Equivalent", then they are considered to implement the same computational algorithm.
  *   <li><b>Numerical Equivalence</b></li> If two Ops are "Numerically Equivalent", then they are considered to output the same values, neglecting trivial differences such as floating-point errors.
  * </ol>
  *
@@ -36,12 +36,11 @@
  *   <li> filter.gauss(ij.ImagePlus, java.lang.Double) -> ij.ImagePlus</li>
  *   <li> filter.gauss(net.imglib2.img.Img, net.imglib2.algorithm.neighborhood.Shape) -> net.imglib2.img.Img</li>
  * </ol>
- * Ops 1 and 2 should be considered form equivalent, as they have the same name, and Op equivalent, as they both take in an image data structure and a floating point number and return an image data structure, but they are likely not numerically equivalent due to the underlying data structures.
+ * Ops 1 and 2 should be considered form equivalent, as they have the same name, and algorithmically equivalent, as they both take in an image data structure and a floating point number and return an image data structure, but they are likely not numerically equivalent due to the underlying data structures.
  * <p>
  * Ops 1 and 3 should also be considered form equivalent, as they have the same name, but are not form equivalent, as one computes its own Shape over which to perform a gaussian blur, while the other takes a predefined shape.
  * <p>
  * These definitions of equivalence provide a level of flexibility impossible without the Ops API; form equivalence allows us to, for example, define equivalent Ops across programming languages and libraries, and then to create scripts that can run unaltered on data types from each of those languages and libraries. Within each form implementation, however, we ensure numeric equivalence and therefore reproducability
- *
  *
  */
 package org.scijava.ops.api;
--- a/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
+++ b/scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
@@ -51,7 +51,12 @@
  * <p>
  * Ops 1 and 3 are also form-equivalent, as they have the same name, but are not structural-equivalent, as one has an implicit Shape over which to perform a gaussian blur, while the other uses an explicitly specified shape.
  * <p>
- * These definitions of equivalence provide a level of flexibility impossible without the Ops API; structural-equivalence allows us to, for example, define equivalent Ops across programming languages and libraries, and then to create scripts that can run unaltered on data types from each of those languages and libraries. Given the same {@link org.scijava.ops.api.OpEnvironment} and inputs, however, we ensure result-equivalence and therefore reproducability
- *
+ * These definitions of equivalence provide clarity when articulating the Ops framework's benefits:
+ * <ol>
+ *   <li>Form-equivalence in Ops allows a comprehensive search of the available algorithms for accomplishing a particular algorithm, by searching its name</li>
+ *   <li>Form-equivalence in data inputs allows us to consider algorithm inputs in a library-agnostic way, making it easier to understand and use algorithms</li>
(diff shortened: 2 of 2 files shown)


scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Ops.java
```diff
@@ -57,6 +57,16 @@ public static <T> RichOp<T> rich(T op) {
 		return (RichOp<T>) op;
 	}
 
+	/**
+	 * Convenience function for getting the {@link OpInfo} of {@code op}
+	 * @param op the Op
+	 * @return the {@link OpInfo} that generated {@code op}
+	 * @param <T> the type of {@code op}
```
ctrueden, line 74: Might want to declare `@throws NullPointerException if the given object is not an op` or some such. (Or does `rich(op)` throw some other exception if it's not a `RichOp`?)

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
+/**
+ * This module contains the API necessary to retrieve, execute, and reason about Ops.
+ * <h1>What is an Op?</h1>
+ *
+ * An Op is an algorithm adhering to the following traits:
+ * <ol>
+ * <li>Ops are stateless - with no internal state, calling an Op two times on the same inputs will produce the same output.</li>
```
ctrueden: I would say "stateless and deterministic"

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
...
+ * <li>Ops are limited to one output - by limiting Ops to one output, we ensure that the algorithms we write are reusable building blocks. Of course, algorithms always need inputs, and there are no limitations on the number of inputs.</li>
```
ctrueden: I would cut this bullet. All Java methods are limited to one output. I don't think it helps the reader to understand ops to dwell on this point.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
...
+ * <li>Ops have at least one name - this name conveys an Op's purpose</li>
```
ctrueden: I would say "Ops have a name" or "Ops are named" rather than "at least one name". Yes, they can have more than one. But the "at least one name" construction is clunky and confusing. Just say they are named.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
+/**
+ * This module contains the API necessary to retrieve, execute, and reason about Ops.
+ * <h1>What is an Op?</h1>
```
ctrueden, line 15: We talked recently about writing up the "philosophy of Ops" or some such, to explain the purpose of the library, design goals, etc. Perhaps this javadoc block would be a good place to put that? There is some material in that direction at https://imagej.net/libs/imagej-ops. I don't love that page though; I think we could do a better job describing the intent of the library from a broader and more easily understood perspective.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
...
+ * <li>Ops are stateless - with no internal state, calling an Op two times on the same inputs will produce the same output.</li>
+ * <li>Ops are limited to one output - by limiting Ops to one output, we ensure that the algorithms we write are reusable building blocks. Of course, algorithms always need inputs, and there are no limitations on the number of inputs.</li>
+ * <li>Ops have at least one name - this name conveys an Op's purpose</li>
+ * </ol>
+ * Using the name and the combination of input and output parameters, we can retrieve, or "match", any Op from within an {@link org.scijava.ops.api.OpEnvironment}. Op calls with the same name and specified inputs/outputs will be reproducible within a particular Op environment.
+ *
+ * <h1>Op Equivalence</h1>
+ * To support the Op matching paradigm, we establish three types of equivalence:
+ * <ol>
+ *   <li><b>Form Equivalence</b></li> Drawn from <a href="https://en.wikipedia.org/wiki/Theory_of_forms">Plato's Theory of Forms</a>, Form Equivalence implies that two objects (which could be Ops, or Op parameters) theoretically draw from the same shared idea (such as an addition operation, or an image, etc.)
```
ctrueden: I thought we were going to remove the Wikipedia link? Anyone familiar with Plato's theory will recognize the correlation, and to anyone else, it doesn't matter.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
...
+ *     <li>For each input position, accept form-equivalent inputs</li>
+ *     <li>Return form-equivalent outputs</li>
+ *   </ol>
+ *   <li><b>Result Equivalence</b></li> Result Equivalence means that {@code o1.equals(o2)} for two outputs {@code o1} and {@code o2} from two Ops.
+ * </ol>
+ * Within the Ops API, each type of equivalence is utilized in the following ways:
+ * <ol>
+ *   <li><b>Form Equivalence</b></li> If two Ops are "Form Equivalent", they are defined under the same name.
+ *   <li><b>Structural Equivalence</b></li> If two Ops are "Structural Equivalent", they share a common form-reduced description, searchable using {@link org.scijava.ops.api.OpEnvironment#descriptions(java.lang.String)}. For example, an Op "math.add" that produces an ImgLib2 Img from Img operands, and another Op "math.add" that produces a NumPy ndarray from ndarray operands, will reduce to a single description "math.add" that produces an image from image operands.
+ *   <li><b>Result Equivalence</b></li> If two Ops are "Result Equivalent", then they are considered to produce equivalent values, using the primary language-specific definition of equality (e.g. `Object.equals`, for Java usage).
```
ctrueden: `are considered to produce equivalent values` &rarr; `produce equal values`

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
...
+ * </ol>
+ *
+ * For example, consider three Ops:
+ * <ol>
+ *   <li> filter.gauss(net.imglib2.img.Img, net.imglib2.type.numeric.real.FloatType) -> net.imglib2.img.Img</li>
+ *   <li> filter.gauss(ij.ImagePlus, java.lang.Double) -> ij.ImagePlus</li>
+ *   <li> filter.gauss(net.imglib2.img.Img, net.imglib2.algorithm.neighborhood.Shape) -> net.imglib2.img.Img</li>
+ * </ol>
+ * Ops 1 and 2 should be considered form-equivalent, as they have the same name, and structural-equivalent, as they both take in an image data structure and a floating point number and return an image data structure, but they are likely not result-equivalent due to the underlying data structures.
```
ctrueden: `should be considered` &rarr; `are`

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
...
+ *     <li>Are Form Equivalent</li>
+ *     <li>Have the same number of inputs and the same number of outputs</li>
+ *     <li>For each input position, accept form-equivalent inputs</li>
+ *     <li>Return form-equivalent outputs</li>
+ *   </ol>
+ *   <li><b>Result Equivalence</b></li> Result Equivalence means that {@code o1.equals(o2)} for two outputs {@code o1} and {@code o2} from two Ops.
+ * </ol>
+ * Within the Ops API, each type of equivalence is utilized in the following ways:
+ * <ol>
+ *   <li><b>Form Equivalence</b></li> If two Ops are "Form Equivalent", they are defined under the same name.
```
ctrueden: `"Form Equivalent"` &rarr; `form-equivalent`

And same for the quoted structural-equivalent. The narrative later uses this unquoted-lower-case-with-dashes way of writing it, and should do the same here, for consistency

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * For example, consider three Ops:
+ * <ol>
+ *   <li> filter.gauss(net.imglib2.img.Img, net.imglib2.type.numeric.real.FloatType) -> net.imglib2.img.Img</li>
+ *   <li> filter.gauss(ij.ImagePlus, java.lang.Double) -> ij.ImagePlus</li>
+ *   <li> filter.gauss(net.imglib2.img.Img, net.imglib2.algorithm.neighborhood.Shape) -> net.imglib2.img.Img</li>
+ * </ol>
+ * Ops 1 and 2 should be considered form-equivalent, as they have the same name, and structural-equivalent, as they both take in an image data structure and a floating point number and return an image data structure, but they are likely not result-equivalent due to the underlying data structures.
+ * <p>
+ * Ops 1 and 3 should also be considered form-equivalent, as they have the same name, but are not structural-equivalent, as one computes its own Shape over which to perform a gaussian blur, while the other takes a predefined shape.
```
ctrueden: `should also be considered` &rarr; `are also`

ctrueden: `one computes its own Shape` &rarr; `one has an implicit Shape`

`the other takes a predefined shape` &rarr; `the other uses an explicitly specified Shape`

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
ctrueden: Does the fact that the number is a floating point number matter for structural equivalence? I.e.: where are we drawing the line with simplification? Number? Real number?

ctrueden: `they are likely not result-equivalent due to the underlying data structures` &rarr; `they may not be result-equivalent, for reasons such as implementation differences between the two algorithms, or precision loss when converting between data structures`

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
```diff
@@ -0,0 +1,46 @@
...
+ * <p>
+ * These definitions of equivalence provide a level of flexibility impossible without the Ops API; structural-equivalence allows us to, for example, define equivalent Ops across programming languages and libraries, and then to create scripts that can run unaltered on data types from each of those languages and libraries. Given the same {@link org.scijava.ops.api.OpEnvironment} and inputs, however, we ensure result-equivalence and therefore reproducability
```
ctrueden: This last paragraph should be rewritten. Here is a suggestion:

> These definitions of equivalence provide clarity when articulating the Ops framework's various benefits:
>
> 1. The [ops simplification mechanism](/link/to/simplification/doc) enables structural-equivalent ops to be collapsed into a common simplified form, usable with appropriate data structures across programming languages and libraries.
> 2. ... [out of time to write more today, sorry, but hopefully you see where I'm going with this]

ctrueden (changes requested):

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/Ops.java
gselzer: Fixed in 29863dab

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: I wrote a little more, aligned with the bullets in https://imagej.net/libs/imagej-ops, in d723b0e4. Let me know what you think @ctrueden

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: Changed in 2417b72c

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: It's been cut in 03c33f39

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: Changed in 03c33f39

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: Removed it in 3349350d

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: Changed in 41a7ee9e

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: See aa8f970c

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: > floating point number matter for structural equivalence? I.e.: where are we drawing the line with simplification? Number? Real number?

I think simplification comes down to "what `simplify` Ops exist for some concrete type" such as `Double` or `DoubleType`

Fixed the other two issues in 4b19d776

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: Fixed in 4b19d77

gselzer: Fixed in c22004f5

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/package-info.java
gselzer: @ctrueden I think this is pretty tricky, but I gave it a go in ed64b87b. Let me know what you think.

PR #95, opened 2023-11-07
Partition Simplification Framework

### LINKED ISSUE #135: Add parameter type information to `ops.descriptions()` output
Take this minimal example where we ask for information on the `"math.subtract"` op in Jython from the Fiji script editor:

```python
# get op environoment
from org.scijava.ops.api import OpEnvironment
ops = OpEnvironment.getEnvironment()

print(ops.descriptions("math.subtract"))
```
This produces the following output:

```python
[math.subtract(
	 Inputs:
		I1 input1
		I2 input2
	 Containers (I/O):
		O result
)
, math.sub(
	 Inputs:
		java.lang.Number input1
		java.lang.Number input2
	 Outputs:
		java.lang.Double output1
)
Aliases: [math.subtract]
]
```

What types are the input (`I1` and `I2`) and output (`O`)? After talking to Gabe, I learned they are `RealTypes`. The output from `ops.descriptions()` does not indicate that these are `RealTypes`. How can we display the parameter types here?

gselzer:
Solving issues like scijava/scijava#146 and acting on scijava/scijava#36 requires us to gain some insight about the forms of types and to coalesce structurally equivalent Ops, and there's a lot of overlap with the simplification framework. Unfortunately, the simplified Ops and types are hidden within the simplification framework.

This PR partitions `SimplifiedOpInfo` into two separate `OpInfo` implementations:
* `SimplifiedOpInfo`s consist of an `OpInfo` along with `focus` Ops to focus the inputs and a `simplify` Op to simplify the output. For example, if the original `OpInfo` is `Double output = foo.bar(Double d1, Double d2)`, the simplified `OpInfo` might be `Number output = foo.bar(Number d1, Number d2)`.
* `FocusedOpInfo` takes a `SimplifiedOpInfo` along with `simplify` Ops to simplify the inputs *to the simple type* and `focus` Ops to focus the output *from the simple type*. To continue the above example, a focused `OpInfo` might be `Integer output = foo.bar(Integer d1, Integer d2)`.

Note that this PR is built atop of #94, to make use of the new `Ops` utility class.


imagej/imagej-ops2/src/test/java/net/imagej/ops2/create/CreateImgTest.java
```diff
@@ -32,9 +32,11 @@
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import java.util.List;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
+import com.google.errorprone.annotations.FormatString;
```
ctrueden: Looks like an unintended import?

scijava/scijava-ops-engine/src/main/java/module-info.java
```diff
@@ -61,6 +92,9 @@
 	    org.scijava.ops.engine.impl.OpCollectionInfoGenerator,
 		org.scijava.ops.engine.matcher.reduce.ReducedOpInfoGenerator;
 
+	provides org.scijava.ops.engine.OpDescriptionGenerator with
```
ctrueden, line 95: Maybe this interface should be part of the SPI module instead?

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
```diff
@@ -91,67 +92,67 @@ public OpBuilder(final OpEnvironment env, final String opName, final Hints hints
...
+         */
+		public List<String> help() {
+		    return env.help(new PartialOpRequest(opName));
+		}
+
+        /**
+         * Finds all Ops matching the current partial OpBuilder request
+         * @return a verbose {@link String} for each Op satisfying the partial request
+         */
+		public List<String> helpVerbose () {
```
ctrueden, line 153: `helpVerbose ()` &rarr; `helpVerbose()` here and everywhere

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpDescription.java
```diff
@@ -105,4 +121,28 @@ private static void appendParam(final StringBuilder sb, final Member<?> arg,
         }
         sb.append("\n");
     }
+
+    public static String typeString(final Type input, final boolean simple) {
```
ctrueden: Can this be private? AFAICT it is only called here in the same class.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpDescriptionGenerator.java
```diff
@@ -0,0 +1,24 @@
+
```
ctrueden, line 29: License header

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -98,9 +101,9 @@
  */
```
ctrueden, line 103: Probably overdue to add `@author Gabriel Selzer` to this class :laughing:

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -146,24 +149,17 @@ public class DefaultOpEnvironment implements OpEnvironment {
...
-	public DefaultOpEnvironment(
-		final Collection<Discoverer> discoverers)
-	{
-		this.discoverers = new ArrayList<>(discoverers);
-		this.manDiscoverer = new ManualDiscoverer();
-		this.discoverers.add(this.manDiscoverer);
 		this.typeService = new DefaultTypeReifier(Discoverer.using(ServiceLoader::load));
 		this.history = OpHistory.getOpHistory();
-		matcher = new DefaultOpMatcher(Discoverer.using(ServiceLoader::load).discover(
+		this.matcher = new DefaultOpMatcher(Discoverer.using(ServiceLoader::load).discover(
```
ctrueden: Ooh, that's a style Pandora's box you're opening! Are you sure you want to do that? Refer to instance fields with `this.` all the time? It would change a lot of code!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -315,32 +315,67 @@ public <T> T bakeLambdaType(T op, Type type) {
...
 		}
 	}
-	
-	@Override
-	public Set<OpInfo> infosFrom(Object o) {
-		return infoGenerators().parallelStream() //
+
+	private List<OpInfo> generateAllInfos(Object o) {
+		return ServiceLoader.load(OpInfoGenerator.class).stream() //
+				.map(ServiceLoader.Provider::get) //
```
ctrueden: Does this relate back to that JPMS visibility hack we discovered before? I think a comment is in order here to explain what the `Object o` here is, what the `ServiceLoader.Provider::get` is doing, etc.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -315,32 +315,67 @@ public <T> T bakeLambdaType(T op, Type type) {
...
 
+	/**
+	 * Helper method to get the descriptions for each {@link OpInfo} in
+	 * {@code infos}
+	 * <p>
+	 * NB we return a {@link List} here to preserve multiple instances of the same
+	 * {@link OpInfo}. This is consistent with {@link DefaultOpEnvironment#infos}
+	 * returning multiple instances of the same {@link OpInfo}. The duplicate
+	 * {@link OpInfo}s are created when Ops have multiple names.
+	 *
```
ctrueden, line 386: Should properly have a `</p>` at the end of this paragraph.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -649,7 +689,6 @@ private RichOp<?> wrap(OpInstance<?> instance, Hints hints) {
 	 * @param type the functional {@link Type} of the {@code op} we're looking for
 	 * @param name the name of the {@code op} we're looking for
 	 * @param typeVarAssigns the mappings of {@link TypeVariable}s to {@link Type}s 
-	 * @return null if the specified type has no functional method
```
ctrueden, line 654: Maybe there should be a `@throws OpMatchingException` explanation of that case? I know, it's private, but still. The javadoc was taken this far! May as well take it all the way.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -682,67 +721,29 @@ private OpRequest inferOpRequest(Type type, String name, Map<TypeVariable<?>, Ty
 		return new DefaultOpRequest(name, type, mappedOutputs[0], mappedInputs);
 	}
 
-	private synchronized void initOpDirectory() {
-		if (opDirectory != null) return;
-		Map<String, List<OpInfo>> tmp = new HashMap<>();
-		// add all OpInfos that are directly discoverable
```
ctrueden, line 690: Wow, this was a dense thicket of behavior. If it were staying, I would demand comments here.

I'm not exactly seeing where this logic went in the new code, though...

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
```diff
@@ -682,67 +721,29 @@ private OpRequest inferOpRequest(Type type, String name, Map<TypeVariable<?>, Ty
...
-	private final BiConsumer<Map<String, List<OpInfo>>, OpInfo> addToOpIndex = (final Map<String, List<OpInfo>> directory, final OpInfo opInfo) -> {
+	private final Consumer<OpInfo> addToOpIndex = (final OpInfo opInfo) -> {
 		if (opInfo.names() == null || opInfo.names().isEmpty()) {
 			log.error("Skipping Op " + opInfo.implementationName() + ":\n" +
 				"Op implementation must provide name.");
 			return;
 		}
 		for (String opName : opInfo.names()) {
-			directory.computeIfAbsent(opName, name -> new ArrayList<>()).add(opInfo);
+			opDirectory.computeIfAbsent(opName, name -> new TreeSet<>()).add(opInfo);
```
ctrueden: Is the use of `TreeSet` so that it always stays sorted?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedInfoTreeGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+import org.scijava.ops.api.InfoTree;
+import org.scijava.ops.api.OpEnvironment;
+import org.scijava.ops.api.Ops;
+import org.scijava.ops.api.RichOp;
+import org.scijava.ops.engine.InfoTreeGenerator;
+import org.scijava.ops.api.OpInfo;
+import org.scijava.types.Nil;
+import org.scijava.types.Types;
+
+public class FocusedInfoTreeGenerator implements InfoTreeGenerator {
```
ctrueden, line 55: javadoc

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedInfoTreeGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+		// Determine the Op type
+		Type[] inTypes = reqSimplifiers.stream().map(op -> Ops.info(op).inputTypes().get(0)).toArray(Type[]::new);
+		Type outType = Ops.info(outputFocuser).outputType();
+		Type type = SimplificationUtils.retypeOpType(Types.raw(simpleInfo.opType()), inTypes, outType);
+
+		OpInfo baseInfo = new FocusedOpInfo((SimplifiedOpInfo) originalChain.info(), type, reqSimplifiers, outputFocuser, copier, env);
+		return new InfoTree(baseInfo, originalChain.dependencies());
+	}
+
+	List<String> parseComponents(String signature) {
```
ctrueden: Should be `private` (as should the version of this function in `SimplifiedInfoTreeGenerator`)

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedOpInfo.java
```diff
@@ -0,0 +1,366 @@
+
```
ctrueden, line 29: license header

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedOpInfo.java
```diff
@@ -0,0 +1,366 @@
...
+import org.scijava.ops.engine.struct.RetypingRequest;
+import org.scijava.priority.Priority;
+import org.scijava.struct.Member;
+import org.scijava.struct.Struct;
+import org.scijava.struct.StructInstance;
+import org.scijava.struct.Structs;
+import org.scijava.types.Nil;
+import org.scijava.types.Types;
+
+public class FocusedOpInfo implements OpInfo {
```
ctrueden, line 79: javadoc

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedOpInfo.java
```diff
@@ -0,0 +1,366 @@
...
+	 * {@code type}). Suppose that no direct match existed, and we tried to find a
+	 * simplified version. This simplified version, because it is a
+	 * Computers.Arity1, would need a {@link Computers.Arity1} copy Op to copy the
+	 * output of the simplified Op back into the preallocated output. But this
+	 * call is already identical to the Op we asked for, and we know that there is
+	 * no direct match, thus we go again into simplification. This thus causes an
+	 * infinite loop (and eventually a {@link StackOverflowError}. This means that
+	 * we cannot find a simplified copy Op <b>unless a direct match can be
+	 * found</b>, at which point we might as well just use the direct match.
+	 * <p>
```
ctrueden, line 300: These paragraphs should be terminated with `</p>`.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedOpInfo.java
```diff
@@ -0,0 +1,366 @@
...
+		double penalty = 0;
+
+		List<Type> originalInputs = simpleInfo.srcInfo().inputTypes();
+		List<Type> inputs = inputTypes();
+		for (int i = 0; i < inputs.size(); i++) {
+			penalty += determineLoss(env, Nil.of(inputs.get(i)), Nil.of(originalInputs
+				.get(i)));
+		}
+
+		// TODO: only calculate the loss once
```
ctrueden: I don't understand this TODO. The following call processes the output/outputType, whereas the above loop processes each input. Does it mean to combine the loop + output calculations into a single function call somehow? Would this make the framework significantly faster? If not, I'd say kill the TODO comment.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/Identity.java
```diff
@@ -35,9 +35,10 @@
...
 import org.scijava.ops.spi.Op;
 import org.scijava.ops.spi.OpClass;
+import org.scijava.priority.Priority;
 
 @OpHints(hints = { Simplification.FORBIDDEN })
-@OpClass(names = "simplify, focus")
+@OpClass(names = "simplify, focus, identity", priority = Priority.LAST)
 public class Identity<T> implements Function<T, T>, Op {
 
 	public Identity() {}
```
ctrueden: This constructor is redundant, no?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/Identity.java
```diff
@@ -35,9 +35,10 @@
 import org.scijava.ops.engine.BaseOpHints.Simplification;
 import org.scijava.ops.spi.Op;
 import org.scijava.ops.spi.OpClass;
+import org.scijava.priority.Priority;
 
 @OpHints(hints = { Simplification.FORBIDDEN })
-@OpClass(names = "simplify, focus")
+@OpClass(names = "simplify, focus, identity", priority = Priority.LAST)
```
ctrueden: Why does this op need to be an `OpClass` rather than a more concise `OpField` using a lambda?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/IdentityLossReporter.java
```diff
@@ -43,16 +44,16 @@
  * @param <T> - the type that is not being simplified.
  */
 @OpHints(hints = {Simplification.FORBIDDEN})
-@OpClass(names = "lossReporter")
-public class IdentityLossReporter<T> implements LossReporter<T, T>, Op {
+@OpClass(names = "lossReporter", priority= Priority.VERY_HIGH)
+public class IdentityLossReporter<U, T extends U> implements LossReporter<T, U>, Op {
```
ctrueden, line 48: Why does this op need to be an `OpClass` rather than a more concise `OpField` using a lambda?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/PrimitiveArraySimplifiers.java
```diff
@@ -44,63 +45,106 @@
...
+			for(int i = 0; i < arr.length; i++) {
+				oa.set(i, arr[i]);
+			}
+			return oa;
+		};
+
+	@OpHints(hints = { Simplification.FORBIDDEN })
 	@OpField(names = "focus")
 	public final Function<ObjectArray<Number>, Byte[]> byteArrFocuser = o -> o
 		.stream().map(b -> b == null ? null : b.byteValue()).toArray(Byte[]::new);
```
ctrueden: Why are we using `Byte[]` rather than `byte[]` here?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/PrimitiveArraySimplifiers.java
```diff
@@ -44,63 +45,106 @@
...
+
+	@OpHints(hints = { Simplification.FORBIDDEN })
 	@OpField(names = "focus")
 	public final Function<ObjectArray<Number>, Integer[]> intArrFocuser = o -> o
 		.stream().map(i -> i == null ? null : i.intValue()).toArray(Integer[]::new);
 
-	@OpHints(hints = {Simplification.FORBIDDEN})
+	@OpHints(hints = { Simplification.FORBIDDEN })
 	@OpField(names = "simplify")
 	public final Function<Short[], ObjectArray<Number>> shortArrSimplifier =
```
ctrueden: Same question: why `Short[]` and not `short[]`?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/PrimitiveArraySimplifiers.java
```diff
@@ -44,63 +45,106 @@
...
+			// TODO: Why doesn't System.arraycopy work?
+			for(int i = 0; i < arr.length; i++) {
+				oa.set(i, arr[i]);
+			}
+			return oa;
+		};
+
+	@OpHints(hints = { Simplification.FORBIDDEN })
 	@OpField(names = "focus")
 	public final Function<ObjectArray<Number>, Double[]> doubleArrFocuser = o -> o
```
ctrueden, line 163: Why `Double[]` and not `double[]`?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedInfoTreeGenerator.java
```diff
@@ -32,54 +32,40 @@
...
-import org.scijava.ops.engine.InfoTreeGenerator;
+import org.scijava.ops.api.OpEnvironment;
 import org.scijava.ops.api.OpInfo;
+import org.scijava.ops.api.Ops;
+import org.scijava.ops.api.RichOp;
+import org.scijava.ops.engine.InfoTreeGenerator;
+import org.scijava.types.Nil;
 
-public class SimplificationInfoTreeGenerator implements InfoTreeGenerator {
+public class SimplifiedInfoTreeGenerator implements InfoTreeGenerator {
```
ctrueden, line 53: I like this rename.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpDescriptionGenerator.java
```diff
@@ -0,0 +1,53 @@
+
```
ctrueden, line 29: license header

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpDescriptionGenerator.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.stream.Collectors;
+
+import org.scijava.ops.api.OpEnvironment;
+import org.scijava.ops.api.OpInfo;
+import org.scijava.ops.api.OpRequest;
+import org.scijava.ops.engine.OpDescription;
+import org.scijava.ops.engine.OpDescriptionGenerator;
+import org.scijava.priority.Priority;
+
+public class SimplifiedOpDescriptionGenerator implements
```
ctrueden, line 49: Javadoc

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpInfo.java
```diff
@@ -134,89 +135,14 @@ public double priority() {
...
-		}
-		catch (OpMatchingException e) {
-			return Double.POSITIVE_INFINITY;
-		}
-	}
-
 	@Override
 	public String implementationName() {
-		return srcInfo.implementationName() + " simplified to a " + opType();
+		return info.implementationName() + "_simple";
```
ctrueden: Instead of `_`, which is a legal character in Java class names, why not an invalid character like `|`, to ensure that there cannot be a name clash between `Foo_simple` the class and `Foo_simple` the simplified `Foo`?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpInfo.java
```diff
@@ -294,29 +226,51 @@ public String version() {
...
 		// original info
 		sb.append(ORIGINAL_INFO);
 		sb.append(srcInfo().id());
 		return sb.toString();
 	}
+
+	private Type[] inTypes(Type[] originalInputs, List<RichOp<Function<?, ?>>> inputFocusers) {
+		Map<TypeVariable<?>, Type> typeAssigns = new HashMap<>();
+		Type[] inTypes = new Type[originalInputs.length];
+		// TODO: This feels kind of inefficient, but we have to do each individually
```
ctrueden: Should this be an `NB:` prefix instead of `TODO:`? It is just explaining why the code is written this way...

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpRequest.java
```diff
@@ -165,20 +172,17 @@ public String getName() {
 
 	@Override
 	public Type getType() {
-		throw new UnsupportedOperationException(
-			"The type of a SimplifiedOpRequest is indeterminate; it must be matched with a OpInfo to form a concrete Type");
+		return type;
```
ctrueden, line 173: :tada:

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpRequest.java
```diff
@@ -191,12 +195,32 @@ public boolean typesMatch(Type opType) {
...
+		if (type == null) return true;
+		if (type instanceof ParameterizedType) {
+			if (!GenericAssignability.checkGenericAssignability(opType,
+					(ParameterizedType) type, typeVarAssigns, true))
+			{
+				return false;
+			}
+		}
+		else {
+			if (!Types.isAssignable(opType, type)) {
```
ctrueden: Could collapse this `if` onto the enclosing `else` to reduce nesting.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/util/Ops.java
```diff
@@ -82,7 +82,7 @@ private Ops() {
 	 * @return
 	 */
 	public static String[] parseOpNames(String names) {
-		return Arrays.stream(names.split(",")).map(s -> s.trim())
+		return Arrays.stream(names.split("\\s*,\\s*")).map(s -> s.trim())
```
ctrueden: Doesn't the `s.trim()` already take care of squelching any whitespace surrounding the commas?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/util/Ops.java
```diff
@@ -162,30 +162,41 @@ public static Class<?> findFunctionalInterface(Class<?> type) {
...
-	 * @return
+	 * @param cls the {@link Class} implmenting some {@link FunctionalInterface}
+	 * @return the functional {@link Method} of {@code cls}
+	 * @throws IllegalArgumentException if {@code cls} does not implement a
+	 *           {@link FunctionalInterface}
 	 */
 	public static Method findFunctionalMethod(Class<?> cls) {
+		Method m = fMethod(cls);
+		if (m == null) {
+			throw new IllegalArgumentException("Op type" + cls +
```
ctrueden: `"Op type" + cls +` &rarr; `"Op type " + cls.getName() +`

(NB: The above suggestion fixes *two* bugs...)

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/util/Ops.java
```diff
@@ -162,30 +162,41 @@ public static Class<?> findFunctionalInterface(Class<?> type) {
...
 		if (iFace == null) {
 			return null;
 		}
-		
-		List<Method> nonDefaults = Arrays.stream(iFace.getMethods())
-				.filter(m -> !m.isDefault()).collect(Collectors.toList());
-		
+
+		List<Method> nonDefaults = Arrays.stream(iFace.getMethods()).filter(m -> !m
+			.isDefault()).collect(Collectors.toList());
```
ctrueden: This change is the victim of autoformatting, which made the line wrapping worse.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -358,6 +359,35 @@ public void testAdaptedOpWithDependencies() {
 	 */
 	@Test
 	public void testSimplificationRecovery() {
+		// Get the Op
+		Computers.Arity1<ObjectArray<Number>, ObjectArray<Number>> c = ops //
+				.op("test" + ".provenanceComputer") //
```
ctrueden, line 363: Why is this string split with the `+`  operator here?

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
```diff
@@ -358,6 +359,35 @@ public void testAdaptedOpWithDependencies() {
...
+
+	/**
+	 * Tests the ability of {@link OpEnvironment#opFromSignature(String, Nil)} to
+	 * generate an Op that has been adapted and simplified.
+	 */
+	@Test
+	public void testFocusedRecovery() {
 		// Get the Op
 		Computers.Arity1<Integer[], Integer[]> c = ops //
 			.op("test" + ".provenanceComputer") //
```
ctrueden: Why is this string split with the `+` operator here?

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/matcher/adapt/OpAdaptationInfoTest.java
```diff
@@ -46,7 +46,7 @@ public class OpAdaptationInfoTest extends AbstractTestEnvironment {
 
 	@BeforeAll
 	public static void addNeededOps() {
-		ops.registerInfosFrom(new CopyOpCollection<>());
+		ops.register(new CopyOpCollection<>());
```
ctrueden, line 49: Every time I see this renamed method, I smile. :smile_cat:

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/matcher/simplify/SimplifyTest.java
```diff
@@ -52,15 +52,17 @@ public class SimplifyTest extends AbstractTestEnvironment implements OpCollectio
...
-		ops.register(new LossReporterWrapper());
-		ops.register(new PrimitiveArraySimplifiers());
-		ops.register(new CopyOpCollection());
-		ops.register(new CreateOpCollection());
+		ops.register( //
+			new SimplifyTest(), //
+			new PrimitiveSimplifiers(), //
+			new PrimitiveLossReporters(), //
+			new IdentityLossReporter(), //
+			new Identity(), //
```
ctrueden: `Identity` should have a `<>`, no?

scijava/scijava-types/src/main/java/org/scijava/types/Types.java
```diff
@@ -1676,6 +1678,10 @@ private static boolean isAssignable(final Type type,
 				return true;
 			}
 
+			if (type.equals(Any.class)) {
```
ctrueden, line 1681: `Any` is not `final`.... so this check would fail if someone extends it.... :wink: 

(To clarify: I think `Any` should probably be final—but if not, this line should change to something like `type instanceof Class && Any.class.isAssignableFrom((Class<?>) type)`.)

scijava/scijava-types/src/main/java/org/scijava/types/Types.java
```diff
@@ -2239,7 +2246,7 @@ private static boolean isAssignable(final Type type,
 				return true;
 			}
 			
-			if (type instanceof Any) {
+			if (type instanceof Any || type.equals(Any.class)) {
```
ctrueden, line 2249: Again, `Any` is not `final`...

ctrueden (commented):
Oof, got through it! I skimmed some parts that were above my level of understanding, but hopefully it helps anyway.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: > Is the use of `TreeSet` so that it always stays sorted?

@ctrueden yes, I want it sorted by Op priority - I'm starting to think that maybe instead of `List<OpInfo>` being returned by `OpEnvironment.infos` we should return a `SortedSet<OpInfo>` - that way, you'd get two guarantees:
1. The sorting gives you Ops by priority
2. The set prevents duplicates.

ctrueden: Yes, `SortedSet` seems like the right interface here! :+1:

imagej/imagej-ops2/src/test/java/net/imagej/ops2/create/CreateImgTest.java
gselzer: Oop, thanks! Removed now

scijava/scijava-ops-engine/src/main/java/module-info.java
gselzer: I can't say I'm a huge fan of that - the goal is for a library like Imglib2-algorithm to extend the SPI, not the API (although of course now we probably just want it to use javadoc, but that's the kind library). I don't see that kind of library also implementing an `OpDescriptionGenerator` - i.e. I think it aligns more with the `MatchingRoutine` interface in the API than the `OpCollection` interface in the SPI.

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java
gselzer: Messed up the template - fixed now!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpDescription.java
gselzer: Fixed!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpDescriptionGenerator.java
gselzer: Added!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: Added

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: Fixed

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: Fixed

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: Added

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: Added it!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: I got rid of it! The `initOpDirectory` method allowed us to lazily initialize the `Map<String, SortedSet<OpInfo>> opDirectory`, but now we eagerly initialize it, so no need for this functionality.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/impl/DefaultOpEnvironment.java
gselzer: Changed things to use `SortedSet` across the board.

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedInfoTreeGenerator.java
gselzer: Added!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedInfoTreeGenerator.java
gselzer: Changed it!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedOpInfo.java
gselzer: Added!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedOpInfo.java
gselzer: Added!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedOpInfo.java
gselzer: Fixed!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/FocusedOpInfo.java
gselzer: Uh, I think the point was that we want to avoid computing the priority every time we call `FocusedOpInfo.getPriority()`, but I have no idea why the `TODO` was there.

Since there isn't a major harm in computing it more than once, I just added a `final Double priority = null` field to `FocusedOpInfo`, and whenever `FocusedOpInfo.getPriority()` is called we compute the priority if `priority == null`. I think that's enough to kill the `TODO`

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/Identity.java
gselzer: Yup, removed it!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/Identity.java
gselzer: I mean, you'll still have the `OpCollection` boilerplate, but just for you I made it into a `OpField`!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/PrimitiveArraySimplifiers.java
gselzer: Uh, no reason, we should have both though, which I have thus added!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/PrimitiveArraySimplifiers.java
gselzer: Uh, no reason, we should have both though, which I have thus added!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/PrimitiveArraySimplifiers.java
gselzer: Uh, no reason, we should have both though, which I have thus added!

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
gselzer: Fixed!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpInfo.java
gselzer: Sure, changed!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpRequest.java
gselzer: Thanks, good idea!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpInfo.java
gselzer: I figured that this would conflict with `SimplifiedOpInfo.signature()`, but I guess it doesn't (which is great!), so I merged it!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpDescriptionGenerator.java
gselzer: Added!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpDescriptionGenerator.java
gselzer: Added

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/util/Ops.java
gselzer: ...it should. I don't remember why I added it in...

Removed it, though.

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/impl/ProvenanceTest.java
gselzer: It isn't anymore!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/util/Ops.java
gselzer: Fixed!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/util/Ops.java
gselzer: Fixed!

scijava/scijava-ops-engine/src/test/java/org/scijava/ops/engine/matcher/simplify/SimplifyTest.java
gselzer: Added it!

scijava/scijava-types/src/main/java/org/scijava/types/Types.java
gselzer: Made `Any` `final`

scijava/scijava-types/src/main/java/org/scijava/types/Types.java
gselzer: It is now!

imagej/imagej-ops2-tutorial/docs/SearchingForOps.md
```diff
@@ -2,13 +2,14 @@
 
 As the `OpEnvironment` is fully extensible, different `OpEnvironment`s might contain different Ops, and it is important to be able to query an `OpEnvironment` about the available Ops.
 
-Users can query the `OpEnvironment` for Ops matching a given name using the method `OpEnvironment.descriptions(String)`. The following SciJava script queries an `OpEnvironment` for all Ops matching the name `filter.gauss`:
+Users can query the `OpEnvironment` for available Ops using the method `OpEnvironment.help(String)`. The following example shows you how to find all `"filter.gauss"` Ops in an `OpEnvironment`:
```
gselzer: We should describe the new namespace searching functionality

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java
```diff
@@ -818,29 +831,63 @@ default void registerInfosFrom(Object o) {
...
+		return helpVerbose(new PartialOpRequest(name));
+	}
+
 
+	/**
+	 * Returns verbose descriptions for all Ops contained within this
+	 * {@link OpEnvironment}
+	 *
+	 * @param request the {@link OpRequest} to filter on
+	 * @return a {@link Set} of descriptions
```
gselzer: These `@return` statements are wrong

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpRequest.java
```diff
@@ -132,4 +135,68 @@ static void append(final StringBuilder sb, final String s) {
...
+	}
+
+	PartialOpRequest(String name, Nil<?>[] args, Nil<?> outType) {
+		this.name = Optional.of(name);
+		Type[] tmp = Arrays.stream(args).map(nil -> nil == null ? null : nil.getType()).toArray(Type[]::new);
+		this.args = Optional.of(tmp);
+		this.outType = Optional.of(outType.getType());
+	}
+
+	@Override public String getName() {
```
gselzer: Could use a format

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpRequest.java
```diff
@@ -132,4 +135,68 @@ static void append(final StringBuilder sb, final String s) {
...
 		sb.append(s);
 	}
+
+}
+
+class PartialOpRequest implements OpRequest {
+
+	private final Optional<String> name;
+	private final Optional<Type[]> args;
+	private final Optional<Type> outType;
```
gselzer: If we're just going to return `null` when these are empty, might as well just set them to `null`...

gselzer (commented):
Started out with a mini-review

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
```diff
@@ -114,6 +114,10 @@ default Type outputType() {
 	default int compareTo(final OpInfo that) {
 		if (this.priority() < that.priority()) return 1;
 		if (this.priority() > that.priority()) return -1;
+		int nameComparison = this.toString().compareTo(that.toString());
```
gselzer: It might be nice to think about this a little more, and also to make sure that it aligns with the `equals` method...

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpDescription.java
```diff
@@ -46,8 +50,8 @@ public final class OpDescription {
      * @param info the {@link OpInfo} of interest
      * @return a descriptor for this {@link OpInfo}
      */
-    public static String basic(final OpInfo info) {
-        return basic(info, null);
+    public static String verbose(final OpInfo info) {
+        return verbose(info, null);
```
gselzer: I kind of wonder if this  class should just be combined with `SimplifiedOpDescriptionGenerator` somehow...

imagej/imagej-ops2-tutorial/docs/SearchingForOps.md
gselzer: First cut in 77539361

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpRequest.java
gselzer: Formatted in 47fd17da

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpEnvironment.java
gselzer: Cleaned up in c407ba2c

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpRequest.java
gselzer: Cleaned up (all of `OpRequest`) in 1b51740e

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpDescription.java
hinerm: @gselzer it seems odd that the `OpDescription` is a static utility class and like it should be able to be rolled into something else. I don't have a sense for how easily this logic would be accessed in the current places using `OpDescription`. If it's trivial then I would be in favor of refactoring.

gselzer: @hinerm right now it's called in two places:
1) Lots of `OpInfo` implementations use it for their `toString()` - it'd be slick to put it in `OpInfo`, but we can only put `default` methods in there, so that's not ideal
2) The `SimplifiedOpDescriptionGenerator` uses it.

One idea is to combine it with either the `Ops` or the `Infos` utility classes...

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
hinerm: @gselzer is simply checking `implementationName` insufficient for some reason? Is it actually possible for the `toString` result of two `OpInfo`s to be the same but the `implementationName` different?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/OpDescription.java
gselzer: Migrated all functionality to `Infos.java` in 1a8f51fc

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
gselzer: @hinerm so the one case where we might want this is [here](https://github.com/scijava/incubator/pull/95/files#diff-af5dec84a4906407c0159a8ff35d0858d018c6a57302bb6e9c8d81d901107e47R77) - that test will fail because [`OpEnvironment.infos()`](https://github.com/scijava/incubator/pull/95/files#diff-b271cbe48149d64e6e0bfe70e70eadb7d5d3e3f0d83988a06c2e837af4831103R174-R176) will consider the two `OpInfo`s to be the same.

What if we instead changed the `List<String> names()` method of `OpInfo` to return a `Set<String>`, and then we could check set equality in `compareTo` and `equals`?

gselzer: @hinerm so the one case where we might want this is [here](https://github.com/scijava/incubator/pull/95/files#diff-af5dec84a4906407c0159a8ff35d0858d018c6a57302bb6e9c8d81d901107e47R77) - that test will fail because [`OpEnvironment.infos()`](https://github.com/scijava/incubator/pull/95/files#diff-b271cbe48149d64e6e0bfe70e70eadb7d5d3e3f0d83988a06c2e837af4831103R174-R176) will consider the two `OpInfo`s to be the same.

What if we instead changed the `List<String> names()` method of `OpInfo` to return a `Set<String>`, and then we could check set equality in `compareTo` and `equals`?

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpDescriptionGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+	@Override
+	public String simpleDescriptions(OpEnvironment env, OpRequest req) {
+		String name = req.getName();
+		if (name == null) {
+			return allNamespaces(env);
+		}
+		if (env.infos(name).isEmpty()) {
+			return allNamespaces(env, name);
+		}
+		var infos = SimplificationMatchingRoutine.getSimpleInfos(env, req.getName());
```
gselzer: Man, I really hate this line. I dislike that we have a separate data structure storing `OpInfo`s that's not in the `DefaultOpEnvironment`, *and* I dislike how the way this is written will ignore all Ops that can't be simplified (which in this case is any `Computer`/`Inplace` whose output can't be copied).

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpDescriptionGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+	private String allNamespaces(final OpEnvironment env, final String name) {
+		List<String> namespaces = env.infos().stream() //
+				// Get all names from each Op
+				.flatMap(info -> info.names().stream()) //
+				// Deduplicate, sort & collect
+				.distinct() //
+				.filter(n -> n.contains(name)) //
+				.sorted() //
+				.collect(Collectors.toList());
+		return "Namespaces:\n\t> " + String.join("\n\t> ", namespaces);
```
gselzer: Maybe this should be `OpNames:`

gselzer (commented):
So close!

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpDescriptionGenerator.java
hinerm: >I dislike how the way this is written will ignore all Ops that can't be simplified (which in this case is any Computer/Inplace whose output can't be copied).

@gselzer this sounds particularly concerning.. what're the implications here? Does that mean that these ops won't show up in an `ops.help` call?

gselzer: > > I dislike how the way this is written will ignore all Ops that can't be simplified (which in this case is any Computer/Inplace whose output can't be copied).
> 
> @gselzer this sounds particularly concerning.. what're the implications here? Does that mean that these ops won't show up in an `ops.help` call?

@hinerm yes, if an Op cannot be simplified then it will not show up in an `OpEnvironment.help` call (but it will still show up in an `OpEnvironment.helpVerbose` call)

gselzer: Fixed (albeit HACKy) with 6bcee354

scijava/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpDescriptionGenerator.java
gselzer: Fixed with 8d9373bd

scijava/scijava-ops-api/src/main/java/org/scijava/ops/api/OpInfo.java
gselzer: Fixed with 82367790

PR #96, opened 2023-11-24
Update Ops discovery docs

### LINKED ISSUE #161: Purge @Plugin annotations
Apparently @ctrueden says that `git grep @Plugin` returns results outside of imagej2-legacy (the only place where it makes sense to see them) - we should remove them all!

ctrueden:
We do not use SciJava Common anymore, nor its `@Plugin` annotation.

Closes scijava/scijava#161.


gselzer (approved):
Looks great! I have no comments!

PR #98, opened 2023-11-30
Migrate Markdown Pages to ReadTheDocs

### LINKED ISSUE #144: Suggest an order for the SciJava Ops docs
The SciJava Ops documentation doesn't provide any strong indication which document should be visited first. I certainly agree that `InstallingSciJavaOpsInFiji.md` makes sense as the entry point, but what about after that? Should I go to `OpBestPractices.md`? `SearchingForOps.md`? In my experience when jumping around in those documents there were concepts I was not yet familiar with so I left those documents thinking I had stumbled into the advanced section. The best document for me after installing SciJava ops was `CallingOps.md` and `ScriptingInImageJ2.md` which broke down the bits I needed to get going.

Adding a TOC (maybe as a README?) or applying numbers to the docs might help here.

### LINKED ISSUE #159: Migrate to ReadTheDocs
Per team discussion today we agreed that we'll make a ReadTheDocs page for our documentation.

This will be our first Java-based RTD page, but we can look at other projects like [QuPath](https://qupath.readthedocs.io/en/0.4/)

gselzer:
This PR migrates all of the Markdown documents currently in `imagej-ops2-tutorial` to a new ReadTheDocs site.

This PR currently only closes scijava/scijava#159

TODO:
* [ ] Create the website
* [ ] Host it on `ops.scijava.org`


gselzer:
Also, scijava/scijava#144 would be closed

PR #106, opened 2023-12-20
Hide internal Op namespaces

### LINKED ISSUE #168: Hide internal ops from ops.help methods
`adapt` and `simplify` aren't meant for direct use and shouldn't be shown in the help results at all.

gselzer:
This PR adds the minimal changes required to hide a set of "internal" Op namespaces from broad `OpEnvironment.help` queries. Note that if you actually search for the namespace, relevant Ops will show up.

Currently, we hide 4 different namespaces:
* `adapt`
* `simplify`
* `focus`
* `lossReporter`

@ctrueden @hinerm please let me know if you are satisfied with these changes. We may have to extend this work in the case that (a) we want this behavior in a DRY way across multiple `OpDescriptionGenerator`s, or (b) we have a hidden namespace that "contains" Ops with a different name - i.e. we want to create some `adapt.foo` Op.

Closes scijava/scijava#168


hinerm:
TODO: move these 4 to an `engine` namespace, and write a regression test to make sure they aren't listed in `help` calls

PR #112, opened 2023-12-29
Add an Ops philosophy

### LINKED ISSUE #143: Create a "SciJava ops philosophy" document
ImageJ ops and SciJava ops clearly have distinct design philosophies that they abide by. While I know some of the reasons for why SciJava ops works the way it does, new users may be really mystified by it. Especially if they have any experience with ImageJ ops. For example when I was first interacting with SciJava ops, I thought the `ops.binary()` was a special namespace for _binary_ images. So this call: `ops.binary("filter.gauss").input(inImage, 2.0).output(outImage).computer()` had me a little confused for a moment.

I think it would be a great starting point for new and existing users (_i.e._ I'd probably would refer to it myself and share it).

gselzer:
This PR adds a new page to the documentation to describe the philosophy of Ops (i.e. "Why Ops").

I'm a bit wary of whether this document makes sense for the audience, so I'm looking for heavy critique from @ctrueden @hinerm @elevans before merging

Closes scijava/scijava#143

### DIFF
--- a/docs/ops/doc/WhyOps.md
+++ b/docs/ops/doc/WhyOps.md
@@ -1,0 +1,46 @@
+# Why Use SciJava Ops?
+
+The fundamental goal of SciJava Ops is to fit the "best" algorithm possible to each task. Historically, identifying and applying the "best" algorithm has been difficult for a variety of reasons:
+* **Technology iterates quickly:** the "best" algorithm five years ago may be replaced by "better" algorithms within new libraries, in different programming languages, incompatible with established workflows.
+* **Hardware dependence limits reuse:** as analysis increasingly migrates to GPU-based calculations, operating environments proliferate and algorithms become less portable.
+* **Algorithm libraries are fragmented:** actually *finding* the "best" algorithm is no simple task, as diverse implementations may exist across any number of programming languages and documentation styles. 
+
+SciJava Ops takes strides to ease these burdens by separating the *what* ("I want to perform a gaussian blur on this image with this sigma value) from the *how* (using scikit-image on zarr arrays). By creating these abstractions, we move towards a single unified, standardized mechanism for applying algorithms. In such an environment, portable workflows can be created quickly and new technologies can be integrated seamlessly.
+
+## What are the driving values of SciJava Ops?
+
+1. **Consistency**:  All Ops are called in the same way, regardless of the mechanisms used by the underlying framework. This means that you don't have to learn Python to call Ops written in Python, but it also means that you could pass the output from an Op written in Python to an Op written in Java, all with the same syntax!
+2. **Reusability**: Ops extends Java's mantra of "write once, run anywhere" to image processing algorithms. Algorithms written for the SciJava Ops framework are usable as-is from any SciJava-compatible software project including ImageJ2/Fiji or from Python using PyImageJ. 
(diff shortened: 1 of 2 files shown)


elevans:
@gselzer So far I like the document and it helps me understand the "gap" that SciJava Ops will fill in the image algorithm space. What I would like to see added to the document at some point is some text that helps soften the blow of the (from my perspective) unusual syntax of a SciJava Op call (_i.e._ `ops.op("filter.gauss").arity2().input(Img, 3.0)`). Specifically, I'm expecting people to really be thrown off on why they need to specify the input arity for their Op call. Some text that says "this is why we do this and why its best" would be great!

gselzer:
@elevans thank you for taking a look.

> What I would like to see added to the document at some point is some text that helps soften the blow of the (from my perspective) unusual syntax of a SciJava Op call (_i.e._ `ops.op("filter.gauss").arity2().input(Img, 3.0)`). Specifically, I'm expecting people to really be thrown off on why they need to specify the input arity for their Op call.

Yeah, that's needed, although we might argue that it is better to put that discussion elsewhere. Is there not something like that in the `Calling Ops` page of the documentation?

>  Some text that says "this is why we do this and why its best" would be great!

I don't think I can write this well, having been an [opponent](https://github.com/scijava/incubator/pull/74#issuecomment-1478357786) of this change. @hinerm or @ctrueden I would really appreciate if one of you could write this section because as proponents of this change you may be able to better convince users that it is worthwhile.

hinerm:
>What I would like to see added to the document at some point is some text that helps soften the blow of the (from my perspective) unusual syntax of a SciJava Op call (i.e. ops.op("filter.gauss").arity2().input(Img, 3.0)

> I don't think I can write this well, having been an [opponent](https://github.com/scijava/incubator/pull/74#issuecomment-1478357786) of this change. @hinerm or @ctrueden I would really appreciate if one of you could write this section because as proponents of this change you may be able to better convince users that it is worthwhile.

You do not *need* to say `ops.op.arity` - you can write `ops.binary`. As I [said here](https://github.com/scijava/incubator/pull/74#issuecomment-1478482715), separating arity and input type to two steps in the builder gives you an additive number of choices to make, and you never see more than `[max arity]` options at a time, while combining them in a single step gives you a multiplicative number of choices.

And as @ctrueden [wrote](https://github.com/scijava/incubator/pull/74#issuecomment-1492391584), this API is Java-developer-focused (e.g. programming from an IDE). We can write convenience layers on top for python/scripting for users, as needed.

I vote against API-level documentation or explanation in this philosophy document.

docs/ops/doc/WhyOps.md
```diff
@@ -0,0 +1,46 @@
...
+* an image and a shape over which to compute the gaussian blur
+* ...
+
+It is SciJava Ops' goal to collapse all algorithms into the following descriptions, and to provide a uniform execution pathway for each - hiding the complexities of each individual algorithm library.
+
+## Isn't SciJava Ops just another algorithm library?
+
+While it may currently seem that way, SciJava Ops is designed to *wrap* existing algorithms, not to create new ones (although you can certainly do that!)
+
+To that end, SciJava Ops
```
hinerm: @gselzer did you have more written that goes here? Do you have more sections planned?

gselzer: @hinerm I just wanted to write something along the lines of "SciJava exposes a rich API for declaring Ops from arbitrary sources, spanning libraries and programming languages". Sorry I forgot!

hinerm (approved):

PR #113, opened 2024-01-08
Refactor descriptions

gselzer:
This PR is (currently) a messy attempt at making the return from `DefaultOpEnvironment.help()` a lot cleaner to read. We make the following broad changes:
* The multiline description is condensed into a single line
* The only name shown is the searched name - no aliases are shown, and I can't think of a reason you'd want to know the aliases for an Op.
* All Op names are required to belong to a namespace. `engine.adapt` Ops now use the `engine.create` and `engine.copy` names where necessary for adaptation purposes. Other Ops were altered as needed to enforce this requirement.
* Argument types are now omitted, and only the names are shown.
* Container and Mutable arguments are now shown using the symbols `*` and `^`, respectively. @ctrueden liked this the best.

Note that this PR is designed to close scijava/scijava#167, and is meant to give @hinerm a starting point since he assigned himself that issue.

Points for further discussion:
* Using the names, and not the types, makes collapsing `OpInfo`s harder, because right now we only collapse based on equal `String`s. But, if the names are scraped from javadoc, we can't stop someone from e.g. making a typo and adding a new entry to the help.
* Do we think that the new key for container and mutable outputs are visible enough?
* ~~I never changed the verbose description - how should that change, if at all?~~ With @elevans' help, I edited the verbose descriptions to look similar to the simple descriptions, but they also include type/implementation/description details.

TODO:
* [ ] Once we finalize the design, update the docs!

PR #118, opened 2024-01-12
Provide more helpful Op matching exceptions

gselzer:
This PR leans on the new `OpEnvironment.help()` API to enable more helpful `OpMatchingException`s.

This means that if the following Op call does not match:
```java
		var list = ops.binary("math.add").input(List.of(5.0), List.of(6.0)).apply();
```

You will now get an error like:
```java
org.scijava.ops.api.OpMatchingException: No match found! Perhaps you meant:
Ops:
	> math.add(
		 Inputs:
			Double input1
			Double input2
		 Outputs:
			Double output1
	)
	
	> math.add(
		 Inputs:
			Number input1
			Number input2
		 Outputs:
			Double output1
	)
	

	at org.scijava.ops.engine/org.scijava.ops.engine.matcher.impl.DefaultOpMatcher.agglomeratedException(DefaultOpMatcher.java:83)
	at org.scijava.ops.engine/org.scijava.ops.engine.matcher.impl.DefaultOpMatcher.match(DefaultOpMatcher.java:74)
	at ...
```
Of course, this output will change with #113, and we might want to make further alterations to the help functionality.

EDIT: Unfortunately, in SciJava Ops Image this change is being bitten by scijava/scijava#176, which means that the build increases by 2 minutes (because of one test file!). For now I'm going to try/catch this issue.

### DIFF
--- a/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/DefaultOpMatcher.java
+++ b/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/DefaultOpMatcher.java
@@ -71,17 +71,24 @@ public class DefaultOpMatcher implements OpMatcher {
 		}
 
 		// in the case of no matches, throw an agglomerated exception
-		throw agglomeratedException(exceptions);
+		throw agglomeratedException(conditions.request(), exceptions, env);
 	}
 
-	private OpMatchingException agglomeratedException(
-		List<OpMatchingException> list)
-	{
-		OpMatchingException agglomerated = new OpMatchingException(
-			"No MatchingRoutine was able to produce a match!");
-		for (int i = 0; i < list.size(); i++) {
-			agglomerated.addSuppressed(list.get(i));
+	private OpMatchingException agglomeratedException( //
+		final OpRequest request, //
+		final List<OpMatchingException> list, //
+		final OpEnvironment env //
+	) {
+		// Develop help message
+		var msg = "No match found!";
+		try {
+			// TODO: Remove try/catch.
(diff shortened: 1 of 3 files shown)


scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/impl/DefaultOpMatcher.java
```diff
@@ -71,17 +71,24 @@ public OpCandidate match(MatchingConditions conditions, OpEnvironment env) {
...
+	private OpMatchingException agglomeratedException( //
+		final OpRequest request, //
+		final List<OpMatchingException> list, //
+		final OpEnvironment env //
+	) {
+		// Develop help message
+		var msg = "No match found!";
+		try {
+			// TODO: Remove try/catch.
+			msg += " Perhaps you meant: \n" + env.help(request);
```
hinerm: conceptually this seems useful. We'll have to see if it is in practice but I don't see any reason not to do it now.

scijava-ops-engine/src/test/java/org/scijava/ops/engine/matcher/OpMatchingExceptionTest.java
```diff
@@ -80,7 +80,7 @@ public void duplicateErrorRegressionTest() {
 		}
 		catch (OpMatchingException e) {
 			Assertions.assertTrue(e.getMessage().startsWith(
-				"No MatchingRoutine was able to produce a match!"));
+				"No match found!"));
```
hinerm, line 83: I prefer the new wording! 👍

hinerm (approved):

PR #119, opened 2024-01-23
Tweak symbols in Op Descriptions

### LINKED ISSUE #178: Tweak how symbols are used in help
Some things to consider:

a) Only printing a key if a particular symbol is actually used, and then only the symbols that are used
b) Not using symbols in `helpVerbose` mode as, with the increase in text, they are much less noticeable than in standard help

If we did `b`, however, we would be losing information.. which seems odd in a "verbose" output

### LINKED ISSUE #179: Update documentation
With https://github.com/scijava/incubator/pull/113 closed, our help output has changed, so docs need updating to reflect these changes

gselzer:
This PR makes slight changes to the default returns of `OpEnvironment.help()` and `OpEnvironment.helpVerbose()`, elimating smaller symbols (and the key describing them) in favor of Java-style annotations. Right now, that only means a `@CONTAINER` mark on any `Computer` output, and a `@MUTABLE` mark on any `Inplace` output.

This PR also consolidates down to two "describe" methods in the `Infos` utility class, as we added two new ones in #113 but did not remove any old ones.

Points of discussion:
* [x] As you can see in the associated test, the verbose descriptions for adapted Op infos are almost unusably long. How should we change them? For example:
```
org.scijava.ops.engine.matcher.adapt.OpAdaptationInfoTest.adaptableOp(java.lang.Double,java.lang.Double)|Adaptor:|Info:org.scijava.ops.engine.adapt.functional.FunctionsToComputers$Function2ToComputer2@0-SNAPSHOT{|Info:org.scijava.ops.engine.copy.CopyOpCollection$copyDoubleArray@0-SNAPSHOT{}}
    > input1 : java.lang.Double
    > input2 : java.lang.Double
    > output1 : @CONTAINER double[]
```

TODO:
* [x] Update documentation

Closes scijava/scijava#178 and scijava/scijava#179

### DIFF
--- a/docs/ops/doc/SearchingForOps.md
+++ b/docs/ops/doc/SearchingForOps.md
@@ -2,7 +2,7 @@
 
 As the `OpEnvironment` is fully extensible, different `OpEnvironment`s might contain different Ops, and it is important to be able to query an `OpEnvironment` about the available Ops.
 
-The `OpEnvironment.help()` API allows you to query an `OpEnvironment` about the types of Ops it contains, as we show in the following sections: 
+The `OpEnvironment.help()` API allows you to query an `OpEnvironment` about the types of Ops it contains, as we show in the following sections. **Note that the example printouts from the help API may not reflect the Ops available in *your* Op environment**. 
 
 ## Searching for operations
 
@@ -19,37 +19,25 @@ This gives the following printout:
 
 ```
 Namespaces:
-	> adapt
 	> coloc
 	> convert
 	> copy
-	> cp
 	> create
 	> deconvolve
-	> eval
+	> expression
 	> features
 	> filter
-	> focus
 	> geom
-	> identify
-	> identity
 	> image
 	> imageMoments
 	> labeling
 	> linalg
 	> logic
-	> lossReporter
 	> map
 	> math
 	> morphology
-	> project
 	> segment
-	> simplify
-	> slice
-	> source
-	> src
 	> stats
-	> test
 	> thread
 	> threshold
 	> topology
@@ -70,10 +58,11 @@ print(ops.help("filter"))
 This gives the following printout:
 
 ```
-Namespaces:
+Names:
 	> filter.DoG
 	> filter.addNoise
 	> filter.addPoissonNoise
+	> filter.addUniformNoise
 	> filter.applyCenterAware
 	> filter.bilateral
 	> filter.convolve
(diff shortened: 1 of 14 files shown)


hinerm:
> * As you can see in the associated test, the verbose descriptions for adapted Op infos are almost unusably long. How should we change them?

You could indent each level of inheritance:
```
org.scijava.ops.engine.matcher.adapt.OpAdaptationInfoTest.adaptableOp(java.lang.Double,java.lang.Double)
      Adaptor: org.scijava.ops.engine.adapt.functional.FunctionsToComputers$Function2ToComputer2@0-SNAPSHOT
            Adaptee: org.scijava.ops.engine.copy.CopyOpCollection$copyDoubleArray@0-SNAPSHOT
```

Other thoughts:
* Does the `Info:` add anything?
* I don't think you need to list the version in the help output since you can't have multiple versions of an op loaded at runtime.
* Does a user ever need to see package names? I can't think of a reason they would, off the top of my head. They are never going to use that information to call the op.
* The parameter info in the op name `(java.lang.Double,java.lang.Double)` is redundant

gselzer:
> > * As you can see in the associated test, the verbose descriptions for adapted Op infos are almost unusably long. How should we change them?
> 
> You could indent each level of inheritance:
> 
> ```
> org.scijava.ops.engine.matcher.adapt.OpAdaptationInfoTest.adaptableOp(java.lang.Double,java.lang.Double)
>       Adaptor: org.scijava.ops.engine.adapt.functional.FunctionsToComputers$Function2ToComputer2@0-SNAPSHOT
>             Adaptee: org.scijava.ops.engine.copy.CopyOpCollection$copyDoubleArray@0-SNAPSHOT
> ```

Well, keep in mind that the "name" used here comes from `OpInfo.implementationName()` currently, which is designed to give users some more information about which Ops exist. Maybe we should change that, though...what would you prefer to see @hinerm? Keep in mind this decision would affect all Ops, not just adapted Ops.

hinerm:
> Well, keep in mind that the "name" used here comes from `OpInfo.implementationName()` currently, which is designed to give users some more information about which Ops exist. Maybe we should change that, though...what would you prefer to see @hinerm? Keep in mind this decision would affect all Ops, not just adapted Ops.

I would consider not changing the `implementationName` method but in verbose output processing looking at replacing `|Info` characters with newlines and indents.

I don't have any other suggestions on shortening the name beyond what I posted above. I fine just leaving it the way it is.

PR #122, opened 2024-02-01
Add additional Computers enabling Op outputs in different spots

### LINKED ISSUE #180: Add computer signatures for output parameters in leading parameter positions
While we have the default assumption that `Computers` have a mutable output parameter as their last argument, this is not true in non-ops libraries. To maximize support it would be nice to have functional interfaces that cover the `Container` param being in alternative positions (much like `Inplaces` now)

- [x] Add new `ArityX_Y` interfaces in `org.scijava.function.Computers`, e.g.:
```
	@FunctionalInterface
	public interface Arity3_2<I1, O, I2, I3> extends Computers.Arity3<I1, I2, I3, O> {
		void compute2(I1 in1, @Container O out, I2 in2, I3 in3);

		@Override
		default void compute(I1 in1, I2 in2, I3 in3, @Container O out) {
			compute2(in1, out, in2, in3);
		}
	}
```

-[ ] Add shorthand signatures (e.g. `COMPUTER3`, `COMPUTER3_1`, `COMPUTER3_2`, `COMPUTER3_3`)

Do we need to update the `Computers` maps?

gselzer:
This PR introduces a couple of (auto-generated) changes used to aid the inclusion of Ops with pre-allocated buffer arguments.
* Introduces a set of `Computers.ArityX_Y` functional interfaces, where `X` is used for *the number of pure inputs* and `Y` is the index (1-indexed) of the output element. Currently, the definition of `X` is chosen to align with the existing `Computers.ArityX` interfaces, and `Y` is chosen to align with the `Inplace` functional interfaces.
* Adds the shorthand syntax for Javadoc type declaration, designed to mimic `Inplace` declaration.

Points of discussion:
* Are the choices for `X` and `Y` the best we could decide upon?
* Should we discourage users from using these interfaces explicitly, instead using `Computers.ArityX`? If so, how?
* What more should we test? Probably optional arguments and adaptation, for starters...


scijava-function/templates/main/java/org/scijava/function/Computers.vm
```diff
@@ -41,15 +41,24 @@ public final class Computers {
 	 * nice to use a BiMap from e.g. Google Guava, but that would require a
```
ctrueden: There is no such thing as an "`i`-th entry" of a map... this statement means that the keys are arity values, right?

scijava-function/templates/main/java/org/scijava/function/Computers.vm
```diff
@@ -41,15 +41,24 @@ public final class Computers {
...
 	 * dependency on that component :(
 	 */
-	public static final HashMap<Integer, Class<?>> ALL_COMPUTERS;
+	public static final HashMap<Integer, HashMap<Integer, Class<?>>> ALL_COMPUTERS;
```
ctrueden: This use of `HashMap` is unnecessary. Better would be `Class<?>[16][16]`, it's simple and directly indexable and more performant in time and space and also thread-safe. And if having nulls outside the triangle bothers you, we could instead do:
```java
ALL_COMPUTERS = new Class<?>[maxArity+1][0];
for (int arity=0; arity<=16; arity++) {
  ALL_COMPUTERS[arity] = new Class<?>[arity];
}
```
or something like that.

scijava-function/src/main/java/org/scijava/function/Computers.java
```diff
@@ -69,46 +41,185 @@ private Computers() {
...
-		ALL_COMPUTERS.put(14, Computers.Arity14.class);
+		ALL_COMPUTERS.get(14).put(14, Computers.Arity14.class);
 		ALL_ARITIES.put(Computers.Arity14.class, 14);
-		ALL_COMPUTERS.put(15, Computers.Arity15.class);
+		ALL_COMPUTERS.get(15).put(15, Computers.Arity15.class);
 		ALL_ARITIES.put(Computers.Arity15.class, 15);
-		ALL_COMPUTERS.put(16, Computers.Arity16.class);
+		ALL_COMPUTERS.get(16).put(16, Computers.Arity16.class);
 		ALL_ARITIES.put(Computers.Arity16.class, 16);
+		ALL_COMPUTERS.get(1).put(0, Computers.Arity1_1.class);
```
ctrueden:
```java
ALL_COMPUTERS[1][1] = Computers.Arity1_1.class;
```
Wouldn't that be nice to read? :smiling_face_with_tear:

scijava-function/src/main/java/org/scijava/function/Computers.java
```diff
@@ -119,7 +230,8 @@ private Computers() {
 	 * @throws NullPointerException If {@code c} is {@code null}.
 	 */
 	public static boolean isComputer(Class<?> c) {
-		return ALL_COMPUTERS.containsValue(c);
+		return ALL_COMPUTERS.values().stream().anyMatch(map -> map.containsValue(
+			c));
```
ctrueden: I think using a `Set` data structure for this would be nicer. You could call it `COMPUTERS` or `COMPUTER_SET`, and then rename `ALL_COMPUTERS` to `COMPUTERS_BY_ARITY` or some such, for clarity.

scijava-ops-engine/src/main/java/org/scijava/ops/engine/yaml/impl/JavaMethodYAMLInfoCreator.java
```diff
@@ -118,10 +118,15 @@ private Class<?> deriveOpType(String identifier, String typeString,
 					" could not be loaded: Inplaces must declare the index of the mutable parameter. For example, if your Inplace is designed to mutate the first argument, please write \"Inplace1\"");
 			}
 		}
-		else if (Pattern.matches("^[Cc]omputer\\s*[0-9]*$", typeString)) {
+		else if (Pattern.matches("^[Cc]omputer$", typeString)) {
 			return Computers.computerOfArity(parameterCount - 1);
 		}
-		else if (Pattern.matches("^[Ff]unction\\s*[0-9]*$", typeString)) {
+		else if (Pattern.matches("^[Cc]omputer\\s*[0-9]*$", typeString)) {
+			Integer a = Integer.parseInt(typeString.substring(typeString.indexOf(
```
ctrueden: Why is this typed on `Integer` rather than `int`?

ctrueden: `typeString.indexOf('r')` is always going to return 7. Why not just write `8` here? If the concern is with magic constants, then `'r'` is a magic constant here too; if we renamed `Computer` to `Crouton` it would also break the logic here, so robustness isn't really a reason to leave it as is.

ctrueden: Also, this logic is buggy, because the string `Computer   9` will match, but throw `NumberFormatException` due to the leading spaces. You could put parentheses around the `[0-9]*` to capture it into a group, and then parse the group as a number.

scijava-function/templates/main/java/org/scijava/function/Inplaces.vm
```diff
@@ -167,5 +168,17 @@ public final class Inplaces {
 		public int mutablePosition() {
 			return mutablePosition;
 		}
+
+		@Override
+		public boolean equals(Object that) {
```
ctrueden, line 173: Should this addition of `equals` to `Inplaces` be done as a separate commit? It doesn't seem relevant to the commit's stated goal.

scijava-function/templates/main/java/org/scijava/function/Inplaces.vm
```diff
@@ -167,5 +168,17 @@ public final class Inplaces {
 		public int mutablePosition() {
 			return mutablePosition;
 		}
+
```
ctrueden, line 171: Ahh, this is where Kotlin would be so nice...
```kotlin
data class InplaceInfo(val arity: Int, val mutablePosition: Int);
```
end of class definition :joy:

ctrueden: Failing that, you could shorten the class by making the fields `public final` instead of `private`, and then dispense with the accessors.

scijava-function/src/main/java/org/scijava/function/Computers.java
```diff
@@ -419,4 +552,2339 @@ default void accept(final I1 in1, final I2 in2, final I3 in3, final I4 in4,
...
+		void compute4(I1 in1, I2 in2, I3 in3, @Container O out, I4 in4);
+
+		@Override
+		default void compute(I1 in1, I2 in2, I3 in3, I4 in4, @Container O out) {
+			compute4(in1, in2, in3, out, in4);
+		}
+	}
+
+	@FunctionalInterface
+	public interface Arity5_1<O, I2, I3, I4, I5, I1> extends
```
ctrueden: This should be `<O, I1, I2, I3, I4, I5>`, shouldn't it? Right now it looks like it just swapped the displaced I param with the O, but it should shunt everything over, no?

ctrueden (changes requested):
Thanks a lot for working on this, @gselzer!

> Are the choices for X and Y the best we could decide upon?

I'm not sure I understand the question, but I'm guessing you mean "Do I agree with what X and Y symbolize?" i.e. X being pure input count and Y being the 1-indexed position of the output, right? If so, yes, I think it's fine. I think it is correct and important that you made Y consistent with the `Inplaces` interface in that regard, and X consistent with the existing `Computers.ArityX` interfaces.

Another way of doing it could be to make Y 0-indexed, and run it up to X inclusive, and get rid of the plain `Computers.ArityX` in favor of `Computers.ArityX_X` which would be equivalent. But... I don't like that as much. The way you did it is good. :+1:

However... I think there is a major bug with the numbering of the inputs. Please double check it. It should always be ordered `I1 in1, I2 in2, I3 in3, ...`, right? But if you look at the current diff, the input displaced by the `O` is always swapped to the end, rather than shunting all the inputs to the right while keeping the order.

> Should we discourage users from using these interfaces explicitly, instead using Computers.ArityX? If so, how?

Nah. They work, and in some cases you need them, like with OpenCV. It'll be enough discouragement just to *not have any examples in the tutorials* about them...

> What more should we test? Probably optional arguments and adaptation, for starters...

I defer to @hinerm on that, since he's the one that will be using these initially.

scijava-function/templates/main/java/org/scijava/function/Computers.vm
gselzer: Updated the Javadoc in 43ae5a4

scijava-function/templates/main/java/org/scijava/function/Computers.vm
gselzer: Updated in 43ae5a4

scijava-function/src/main/java/org/scijava/function/Computers.java
gselzer: Updated in 43ae5a4

scijava-function/src/main/java/org/scijava/function/Computers.java
gselzer: I just use `ALL_ARITIES` now for this

scijava-function/src/main/java/org/scijava/function/Computers.java
gselzer: Fixed in e843de3

scijava-function/templates/main/java/org/scijava/function/Inplaces.vm
gselzer: > ```kotlin
> data class InplaceInfo(val arity: Int, val mutablePosition: Int);
> ```

Well, if we used Java 16+, we could get nearly the same syntax with records!

> Failing that, you could shorten the class by making the fields `public final` instead of `private`, and then dispense with the accessors.

Sure, but it's also not a change that I feel like we *need* to make, so not gonna do it here :wink:

scijava-function/templates/main/java/org/scijava/function/Inplaces.vm
gselzer: Probably - made it a separate commit (6d5ec93)

scijava-ops-engine/src/main/java/org/scijava/ops/engine/yaml/impl/JavaMethodYAMLInfoCreator.java
gselzer: That logic should be fixed now!

hinerm (approved):

hinerm:
>What more should we test? Probably optional arguments and adaptation, for starters...

I don't know have other suggestions right now.. maybe once I'm using them?

PR #123, opened 2024-02-07
Add OpenCV ops

### LINKED ISSUE #158: Generate an `ops.yaml` for a 3rd-party library
I believe that external configuration for defining Ops is essential functionality. I would love to see an `ops.yaml` use-case showing inclusion of some other algorithm library into Ops.

I have no personal preferences for what that library would be; something small and simple is fine. Perhaps @elevans has a suggestion of something he'd like to see?

The general procedure would be [as described here](https://github.com/scijava/scijava/issues/56#issuecomment-1815009117) - though I would prefer if the `ops.yaml` was auto-generated

Current development on [this branch](https://github.com/scijava/incubator/tree/opencv-ops)

hinerm:
Hooray it's an external library wrapped into Ops with no code.

Closes https://github.com/scijava/scijava/issues/158

Questions/notes:
- [x] Currently I'm doing nothing with `priority` of the wrapped ops. Thoughts?
- [x] Missing out on the parameter names/documentation 😞 
- [x] My `Computer` heuristic is stupid and tailored to OpenCV. I saw it as something we could expand support for as needed, but wanted to mention it.
- [x] Are there other methods in OpenCV you'd like to see wrapped?
- [x] Any suggestions for changes to the aliases I picked? I expect I missed existing Ops when I was naming some of the wrappers.
- [x] Deciding what to wrap was overwhelming. I ended up *only* using a curated "white" list in the YAML. Wrapping everything, even with a block list, makes me nervous about getting tons of junk. Also any auto-wrapped ops won't have "Ops-style" aliases. So as I was assessing what I *did* want to wrap and what their aliases should be, it felt unnecessary to also be documenting a "block" list. So I didn't implement a block list for OpenCV since I wasn't going to use it. If we end up wanting it, it would be trivial to add another mode to the parser to just grab all the methods in a class, with a block list. Any objections?

Remaining tasks:
- [x] Wrap the ImageJ-opencv conversion into Ops


scijava-ops-ext-parser/README.md
```diff
@@ -0,0 +1,59 @@
...
+  - ...
+
+# Optional. If "containers" is present, the value is a list of fully qualified
+# class names. Op methods that contain 2 or more of parameters with these types
+# will be marked as Computers, with the second occurrance being the "output"
+# param.
+containers:
+  - "container.1.name"
+  - "container.2.name"
+  - ...
```
gselzer: This could be slick, but reflecting upon the current scijava-image, I have a feeling that this design will break down in multiple ways, as for example we have:
*  An [`image.fill`](https://github.com/scijava/incubator/blob/0c1af14e19981fd7ce03bd07ecf4e544c8b6ee03/scijava-ops-image/src/main/java/org/scijava/ops/image/image/fill/Fills.java#L53) computer Op that only has one instance of an image parameter
* A [`coloc.pValue`](https://github.com/scijava/incubator/blob/0c1af14e19981fd7ce03bd07ecf4e544c8b6ee03/scijava-ops-image/src/main/java/org/scijava/ops/image/coloc/pValue/DefaultPValue.java#L81) Op that takes two `RAI`s but is *not* a computer.

> \# Op methods that contain 2 or more of parameters with these types
\# will be marked as Computers, with the second occurrance being the "output"
\# param.

I assume you meant the final occurrence is the output parameter? Do you think that assumption will hold up?

scijava-ops-ext-parser/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+# You can include as many base classes as you wish. Each fully-qualified class
+# name should map to a map of method names to aliases. All methods in the base
+# class with that name will be indexed under the given alias. The alias should
+# be "SciJava" style - e.g. if your method performs a Gaussian filter, alias it
+# "filter.gauss"
+fully.qualified.className:
+  method1Name: "method1.alias"
+  method2Name: "method2.alias"
+  ...
```
gselzer: If the goal is to avoid the pitfalls I mentioned above, then an alternative design that would be more explicit, at the cost of adding more lines, would be:
```yaml
fully.qualified.className:
  - method1Name:
    type: Computer
    aka: "method1.alias"
  - method2Name: # Assumed to be a function?
    aka: "method2.alias"
```
(we can also debate the method entry parameters, e.g. `aliases` vs. `aka` - but we may want to ensure multiple aliases in YAML format)

scijava-ops-ext-parser/README.md
```diff
@@ -0,0 +1,59 @@
+# SciJava Ops External Parser
+
+This module provides a utility that can write Op YAML based on external
+configuration (itself a YAML). This allows inclusion of static methods from
```
gselzer, line 4: This is super awesome - can you update the docs to describe this, in a similar way?

scijava-ops-ext-parser/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+			<plugin>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<!-- Avoid annotation processing here-->
+				<!-- We don't want to error by running our own processor-->
+				<configuration combine.self="override">
+					<proc>none</proc>
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
```
gselzer: Probably not necessary here

scijava-ops-ext-parser/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+			<groupId>org.bytedeco</groupId>
+			<artifactId>openblas</artifactId>
+			<scope>runtime</scope>
+			<classifier>${scijava.natives.classifier.opencv}</classifier>
+		</dependency>
+		<dependency>
+			<groupId>org.bytedeco</groupId>
+			<artifactId>opencv</artifactId>
+			<classifier>${scijava.natives.classifier.opencv}</classifier>
+			<scope>runtime</scope>
```
gselzer: Is all of this necessary for the parser itself? I'd really prefer instead commenting out an example block, something like:

```xml
<!--		<dependency>-->
<!--			<groupId>your.groupId</groupId>-->
<!--			<artifactId>your.artifactId</artifactId>-->
<!--			<version>your.version</version>-->
<!--			<scope>runtime</scope>-->
<!--		</dependency>-->

```

scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplifiedOpRequest.java
```diff
@@ -101,8 +103,14 @@ public SimplifiedOpRequest(OpRequest req, OpEnvironment env) {
...
-		this.outputCopier = mutableIndex == -1 ? null : simplifierCopyOp(env, req
-			.getArgs()[mutableIndex]);
+		if (mutableIndex >= 0 && !(isIdentity(inSimplifiers.get(mutableIndex)) &&
+			isIdentity(outFocuser)))
+		{
+			this.outputCopier = simplifierCopyOp(env, req.getArgs()[mutableIndex]);
+		}
+		else {
+			this.outputCopier = null;
+		}
```
gselzer, line 116: Can we write a test for this? i.e. match a simplified Op with identity chains for the mutable/output type, and then check to make sure there's no copier?

scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/simplify/SimplificationUtils.java
```diff
@@ -215,6 +217,17 @@ public static SimplifiedOpInfo simplifyInfo(OpEnvironment env, OpInfo info) {
...
+	/**
+	 * Helper method to check if an op is an "identity" op - i.e. a naive wrapper
+	 * that does no modification to its base op.
+	 *
+	 * @param op {@link RichOp} to check
+	 * @return {@code true} iff the op is an identity op
+	 */
+	public static boolean isIdentity(RichOp<?> op) {
+		return op.infoTree().info().names().contains("engine.identity");
+	}
```
gselzer: You could use `org.scijava.ops.api.Ops.info(op)` - might be simpler.

scijava-ops-ext-parser/README.md
```diff
@@ -0,0 +1,59 @@
...
+  - "container.2.name"
+  - ...
+
+# You can include as many base classes as you wish. Each fully-qualified class
+# name should map to a map of method names to aliases. All methods in the base
+# class with that name will be indexed under the given alias. The alias should
+# be "SciJava" style - e.g. if your method performs a Gaussian filter, alias it
+# "filter.gauss"
+fully.qualified.className:
+  method1Name: "method1.alias"
```
gselzer: So the method `fully.qualified.className.method1Name` will be named `"ns.method1Name"` *and* `"method1.alias"`? A note somewhere about this would be helpful

scijava-ops-ext-parser/src/main/java/org/scijava/ops/parser/OpParser.java
```diff
@@ -0,0 +1,250 @@
...
+	{
+		String opSource = "javaMethod:/" //
+			+ className //
+			+ "." //
+			+ methodName //
+			+ "%28" //
+			+ Arrays.stream(parameterTypes).map(Class::getName).collect(Collectors
+				.joining("%2C")) //
+			+ "%29";
+		return opSource;
```
gselzer: Is there a reason you didn't use `URLEncoder.encode` here?

scijava-ops-opencv/opencv-ops.yaml
```diff
@@ -0,0 +1,78 @@
+namespace: "cv"
+version: "1"
```
gselzer, line 2: We never talked about a version syntax for Ops - is this how we want to do it? We should probably have documentation somewhere on how versioning should work for Ops.

scijava-ops-opencv/src/test/java/org/scijava/ops/opencv/TestOpenCV.java
```diff
@@ -0,0 +1,199 @@
...
+
+	/**
+	 * Verify we get all the expected ops for {@link #TEST_OP}
+	 */
+	@Test
+	public void testDiscovery() {
+		final OpEnvironment ops = OpEnvironment.build();
+		assertEquals(EXPECTED_HELP, ops.help(TEST_OP));
+		assertEquals(EXPECTED_HELP_VERBOSE, ops.helpVerbose(TEST_OP));
+	}
```
gselzer, line 138: What do you think about a regression test somewhere to show that we actually can get all of the X ops that we purport to expose?

gselzer (commented):
Thanks so much for all this work @hinerm!

Couple of additional comments I have:
* I think it is important to add some mechanism to add descriptions to parameters. Ideally, we'd be able to scrape the javadoc, but that could be tricky.
* ~~How much additional effort would it require to add some simplifiers on top of this, so that you can call OpenCV functions with `RAI`s?~~ (I see you mentioned this as a TODO, which is great)

scijava-ops-opencv/opencv-ops.yaml
```diff
@@ -0,0 +1,78 @@
...
+  pyrDown: "image.pyrdown"
+  pyrUp: "image.pyrup"
+  buildPyramid: "image.pyrbuild"
+  calcHist: "image.histogram"
+  calcBlackProject: "image.blackProject"
+  equalizeHist: "image.equalize"
+  watershed: "image.watershed"
+  pyrMeanShiftFiltering: "image.pyrshift"
+  grabCut: "image.grabCut"
+  distanceTransform: "segment.distTransform"
```
gselzer: SciJava Ops Image has `morphology.distanceTransform`

scijava-ops-opencv/opencv-ops.yaml
```diff
@@ -0,0 +1,78 @@
...
+  distanceTransformWithLabels: "segment.distTransformLabeled"
+  floodFill: "morphology.floodFill"
+  blendLinear: "image.blend"
+  cvtColor: "image.colorConvert"
+  cvtColorTwoPlane: "image.colorConvertTwoPlane"
+  demosaicing: "image.demosaic"
+  HuMoments: "image.huMoment"
```
gselzer: SciJava Ops has an `imageMoments` namespace for Hu Moments

gselzer:
> Questions/notes:
> 
>     * [ ]  Currently I'm doing nothing with `priority` of the wrapped ops. Thoughts?

I think that the only thing you *can* do is enable users to set a priority in your allowlist...

>     * [ ]  Missing out on the parameter names/documentation 😞

As far as names go, if you're accessing the methods via reflection, then you should have access to the `Parameter` objects, and you can call `Parameter.getName()`.

Like I mentioned in my review, I think that short of actually parsing the `.java` files, you won't be able to get the descriptions. You could allow for copying and pasting of descriptions in your YAML, or you could change the parser radically to run on `.java` files, which maybe would be best accomplish later with an issue filed for now.

>     * [ ]  My `Computer` heuristic is stupid and tailored to OpenCV. I saw it as something we could expand support for as needed, but wanted to mention it.

Yeah, like I mentioned in my review I have concerns that this will hold up.

>     * [ ]  Are there other methods in OpenCV you'd like to see wrapped?

The [array operations](https://docs.opencv.org/4.x/d2/de8/group__core__array.html#gadd6cf9baf2b8b704a11b5f04aaf4f39d) may be useful, but we could also add more later.

>     * [ ]  Any suggestions for changes to the aliases I picked? I expect I missed existing Ops when I was naming some of the wrappers.

Just added in some comments where I found aliases in SciJava Ops that differed.

>     * [ ]  Deciding what to wrap was overwhelming. I ended up _only_ using a curated "white" list in the YAML. Wrapping everything, even with a block list, makes me nervous about getting tons of junk. Also any auto-wrapped ops won't have "Ops-style" aliases. So as I was assessing what I _did_ want to wrap and what their aliases should be, it felt unnecessary to also be documenting a "block" list. So I didn't implement a block list for OpenCV since I wasn't going to use it. If we end up wanting it, it would be trivial to add another mode to the parser to just grab all the methods in a class, with a block list. Any objections?

None here!

gselzer:
One final comment - have you thought about the case where there are multiple methods of the same name (i.e. they have different parameters) - your YAML allowlist schema may want to have a way to specify some but not others.

scijava-ops-ext-parser/README.md
hinerm: Like I said, I have no presumption that this heuristic will hold generally, but it seems to be true for OpenCV. If we have other use cases in the future that necessitate generalization I would do so then.

scijava-ops-opencv/opencv-ops.yaml
hinerm: I don't see a reason for that to impact this PR. I split to a [new issue](https://github.com/scijava/scijava/issues/185)

hinerm:
>have you thought about the case where there are multiple methods of the same name 

Yeah this was the intent, to avoid having to explicitly include each method.. if you're making decisions on each method individually, I would consider just writing the `yaml` by hand.

>then an alternative design that would be more explicit, at the cost of adding more lines, would be:

Yeah I think I like this, and we can just say `Computer2` in the config yaml to indicate the container position, and can throw priority and author in there as well (with global overrides)

hinerm:
@gselzer do you mind taking a quick look to see if anything glaring stands out? it does not need to be exhaustive.

scijava-ops-ext-parser/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+			<scope>test</scope>
+		</dependency>
+
+		<!-- Dependencies for OpenCV -->
+		<!--
+		<dependency>
```
gselzer, line 135: Maybe get rid of this?

scijava-ops-opencv/pom.xml
```diff
@@ -0,0 +1,330 @@
...
+			<roles>
+				<role>founder</role>
+				<role>lead</role>
+				<role>developer</role>
+				<role>debugger</role>
+				<role>reviewer</role>
+				<role>support</role>
+				<role>maintainer</role>
+			</roles>
+		</developer>
```
gselzer: Lots of people here aren't developers or contributors

scijava-ops-opencv/pom.xml
```diff
@@ -0,0 +1,330 @@
...
+		<dependency>
+			<groupId>org.scijava</groupId>
+			<artifactId>scijava-ops-api</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.scijava</groupId>
+			<artifactId>scijava-ops-spi</artifactId>
+			<version>${project.version}</version>
+		</dependency>
```
gselzer: Are all of these dependencies needed? If so, probably only at test/runtime?

gselzer (commented):
Just a couple things I saw in the POMs that could be cleaned up!

PR #125, opened 2024-02-09
Port FlimJ-Ops to SciJava Ops

### LINKED ISSUE #205: Parallelize SciJava Ops FLIM Ops
The FLIM Ops contained within [flimj-ops](https://github.com/flimlib/flimj-ops) used [ChunkerOp](https://github.com/flimlib/flimj-ops/blob/efd6d84ea67652b0ff1559a36cb7c610d3773e5b/src/main/java/flimlib/flimj/fitworker/AbstractSingleFitWorker.java#L117), which is unavailable within SciJava Ops FLIM. For the time being, I've made the computation single-threaded, and added a `TODO` in the code, but at some point it may be nice to restore multi-threaded computation.

gselzer:
This PR adds a new module, SciJava Ops Flim, which ports the (ImageJ) Ops from [FlimJ Ops](https://github.com/flimlib/flimj-ops).

TODO:
* [x] I had an issue encapsulating the SciJava Ops Flim code within a module, which I believe stems from native code access. Currently, the module info exists but is disabled through renaming, however we should either restore it or delete it (in its own commit, so we can find it again if necessary).
* [x] I removed parallel computation from `AbstractSingleFitWorker` because the routine depends on the `Chunk` class, which is internal to SciJava Ops Image. Do we care to add parallel computation back in? Should we add a compile-time dependency on SciJava Ops Image to enable us to depend on `Chunk`, or do something else? *EDIT: I filed scijava/scijava#205 for this.*
* [x] I did not port [this notebook](https://github.com/flimlib/flimj-ops/blob/master/notebooks/FlimLib_from_python.ipynb) describing FlimJ Ops access from Python, as those mechanisms are not yet set in stone. Maybe we should file a Paper 3 issue to port that notebook?  *EDIT: I filed scijava/scijava#206 for this.*
* [x] I cannot reproduce [this image](https://github.com/flimlib/flimj-ops/blob/master/images/example%20tau.png), using **either** the groovy script in the [FlimJ Ops README](https://github.com/flimlib/flimj-ops?tab=readme-ov-file#example-usage), or with the new script in the updated README in this PR. Should we update the image below, which I observe running both scripts? *EDIT: I incorrectly contrasted the image, thanks to @elevans for figuring that out. Added a note in the readme that hopefully prevents others from encountering this issue.*
image


gselzer:
Closes scijava/scijava#205

PR #127, opened 2024-02-19
Refactor TypeReifier to Check Against a Map

gselzer:
This PR makes a rather drastic change to the `TypeReifier` logic, necessary to enable the reification of `Dataset` objects into `ImgPlus<T>` types. These changes pass all tests.

This issue arises from the fact that `Dataset` doesn't have any type variables, making type reification end immediately right [here](https://github.com/scijava/incubator/blob/b25b35682f9e6e79581e832e92a67f633e9e2777/scijava-types/src/main/java/org/scijava/types/TypeReifier.java#L116). And, if we simply removed that check, it would immediately break out of the following `for` loop because [`numVars is zero`](https://github.com/scijava/incubator/blob/b25b35682f9e6e79581e832e92a67f633e9e2777/scijava-types/src/main/java/org/scijava/types/TypeReifier.java#L133).

To solve this issue, I refactored the set of `TypeExtractor` plugins to provide a `baseClass` - `TypeExtractor`s are then hashed on this baseclass. Because the javadoc of `TypeToken.getTypes()` always lists subtypes before supertypes, prioritization should still allow narrower `TypeExtractor`s to override broader ones.

The result of this change is that we can allow arbitrary "reification" of types, meaning that we can map `Dataset` to e.g. `ImgPlus<DoubleType>` even if that type *isn't in the type hierarchy* of `Dataset`.

Obviously, it's a big change, and I'm seeking review.


scijava-types/src/main/java/org/scijava/types/DefaultTypeReifier.java
```diff
@@ -67,9 +63,13 @@ private synchronized void initExtractors() {
...
-	private List<TypeExtractor> getInstances() {
-		return discoverers.parallelStream() //
+	private Map<Class<?>, TypeExtractor> getInstances() {
+		Map<Class<?>, TypeExtractor> instances = new HashMap<>();
+		discoverers.stream() //
 			.flatMap(d -> d.discover(TypeExtractor.class).stream()) //
-			.sorted().collect(Collectors.toList());
+			// Find the correct set for each extractor and add it.
+			.sorted() //
+			.forEach(e -> instances.putIfAbsent(e.baseClass(), e));
```
ctrueden: This makes it so that if there are two `TypeExtractor` plugins with the same base class, the first one *silently* wins. It would be better—if we are certain that there should never be two `TypeExtractor` plugins operating on the same base class—to issue an error or at least a warning when a second one is present.

scijava-types/src/main/java/org/scijava/types/SubTypeExtractor.java
```diff
@@ -41,37 +41,31 @@
...
-		return getRawType().isAssignableFrom(cls);
-	}
-
 	@Override
 	public Type reify(final TypeReifier r, final Object object) {
-		if (!canReify(r, object.getClass())) throw new IllegalArgumentException(
-			this + " can only reify Objects of Class " + getRawType()
-				.getSimpleName() + "!");
+		if (!baseClass().isAssignableFrom(object.getClass()))
+			throw new IllegalArgumentException(this +
```
ctrueden: Instead of `this +`, should this error message be `this.getClass().getName() +`? I know it's just left alone from how it was before, but do these type extractor plugins actually stringify in the desired way for this error message to read correctly? I'm guessing not...

scijava-types/src/main/java/org/scijava/types/TypeReifier.java
```diff
@@ -110,56 +110,24 @@ default Type reify(final Object o) {
...
-					if (vars.containsKey(typeVar)) {
-						resolved.putIfAbsent(typeVar, vars.get(typeVar));
-					}
-				}
+		for (final var token : TypeToken.of(c).getTypes()) {
+			Optional<TypeExtractor> opt = getExtractor(token.getRawType());
+			if (opt.isPresent()) {
+				return opt.get().reify(this, o);
 			}
 		}
```
ctrueden, line 129: Some comments would be appreciated for this new code. They should be at least as witty as my comments that got deleted above. :stuck_out_tongue_closed_eyes:

ctrueden (commented):
This looks like a very nice change, in that it simplifies the type extraction logic, using fewer lines of code. It is less flexible than the old design. But in practice, since all current tests pass with this new approach, probably flexible enough for the actual requirements of extensible generic type reasoning at runtime, eh?

Before merging, I would think very carefully about the doors we are closing. For example: what if two type extractors both work on the same base type interface, but for different generics cases? Can we ever imagine such a design being needed?

I don't have any concrete objections to the new design—just a little hesitation about whether it boxes us into a corner later. But API-wise, the only change is making the (renamed) `baseClass()` method public instead of protected. Is it necessary to expose that method? Because if we ever wanted to broaden the design in the future, such that a type extractor might work on multiple base classes, this method would become awkward to keep around only for backwards compatibility.

scijava-ops-legacy/src/test/java/org/scijava/legacy/types/DatasetTypeExtractorTest.java
```diff
@@ -65,6 +67,11 @@ public void testDatasetTypeExtractor() {
 		var actual = new DatasetTypeExtractor().reify(reifier, ds);
 		var expected = Types.parameterize(ImgPlus.class, new Type[] { new Any() });
 		Assertions.assertEquals(expected, actual);
+		OpEnvironment env = OpEnvironment.build();
```
ctrueden, line 76: Ahh, there is an `NB` comment above that is actually obsolete, isn't it?

gselzer:
> This looks like a very nice change, in that it simplifies the type extraction logic, using fewer lines of code. It is less flexible than the old design. But in practice, since all current tests pass with this new approach, probably flexible enough for the actual requirements of extensible generic type reasoning at runtime, eh?

Yeah, the decrease in flexibility was something I too was concerned with, and it is the reason I wanted review. I agree that all tests passing is a promising aspect.
 
> Before merging, I would think very carefully about the doors we are closing. For example: what if two type extractors both work on the same base type interface, but for different generics cases? Can we ever imagine such a design being needed?

Suppose two type extractors work on the same base class `BaseClass<T1, T2, ..., TN>`. In that case, they have access to the same methods on the object they reify - this means either `Ti` can either be determined by calling some method and reifying the result recursively (which is what `TypeExtractors` are supposed to do), or, if no method returns an object of type `Ti`, it is substituted with an `Any`. I don't see how two different type extractors can thus return different answers unless one is written without taking full advantage of that interface's API.

A case that concerns me more is that we lose the ability to use multiple reifiers to reify a single type. Suppose, for example, we have a type `FooBar<A, B> implements Foo<A>, Bar<B>`. If we can reify `Foo`s and `Bar`s, then the old design would give us a `FooBar<A, B>`, where the new design would give us a `Foo<A>` or a `Bar<B>`.

The solution, though, is to write a `FooBarTypeExtractor` that works on `FooBar`s, with (unfortunately) duplicate logic from each of the two simpler extractors. Because `TypeToken.of(c).getTypes()` is *documented* to return Types with subtypes before supertypes, we'll thus hit the `FooBar` type *before* `Foo` or `Bar`, and we should use the correct reifier.

> But API-wise, the only change is making the (renamed) `baseClass()` method public instead of protected. Is it necessary to expose that method? Because if we ever wanted to broaden the design in the future, such that a type extractor might work on multiple base classes, this method would become awkward to keep around only for backwards compatibility.

Well, it's not *really* a rename, it's more a broader usage. I realized that the method `SubTypeExtractor.getRawType()` was needed beyond `SubTypeExtractor` for the map in `DefaultTypeReifier`, so I got rid of the method in `SubTypeExtractor`. And, because it's an interface method, it has to be public, which I agree is unfortunate.

I think that if it worked on multiple classes, we should make multiple type extractors. It makes the complexity within `TypeReifier` much more manageable.

scijava-types/src/main/java/org/scijava/types/DefaultTypeReifier.java
gselzer: Good catch - I warn now in e60e711b

scijava-types/src/main/java/org/scijava/types/SubTypeExtractor.java
gselzer: Changed in 2e55efac

scijava-types/src/main/java/org/scijava/types/TypeReifier.java
gselzer: Added in 11cc558e

scijava-ops-legacy/src/test/java/org/scijava/legacy/types/DatasetTypeExtractorTest.java
gselzer: Heh, it's complicated, but it still holds. JPMS shenanigans.

The `scijava-ops-legacy` module does not `use` `TypeExtractor`, so no TypeExtractors are found using the Discoverers in the declaration of the `DefaultTypeReifier`. Therefore, when we *directly* call the `DatasetTypeExtractor`, the recursive call to reify the `ImgPlus` finds no extractors and returns `ImgPlus<Any>`.

In the line you linked, the `TypeReifier` is constructed within the SciJava Ops Engine, whose module *does* `use` the `TypeExtractor` implementations, so it can find all of the `TypeExtractor` implementations, and we get back the correct type. Note that `scijava-ops-legacy` does *not* depend on scijava-ops-image, so the reifier actually being used here is the very slow `IterableTypeExtractor`.

I described the situation in 449f98f1

PR #130, opened 2024-02-28
Progress enhancements

gselzer:
This PR enables a few enhancements designed to provide Ops more control over the way they report progress, namely:
* Providing access to the current `Task` object, allowing Ops to save it. This can be necessary where the `ThreadLocal` paradigm breaks down, namely within `LoopBuilder` calls.
* By default ignoring the progress updates of `OpDependency`, unless the Op developer wishes for them to be forwarded to the user.

### DIFF
--- a/scijava-progress/src/main/java/org/scijava/progress/Progress.java
+++ b/scijava-progress/src/main/java/org/scijava/progress/Progress.java
@@ -67,14 +67,12 @@ public final class Progress {
 	 * A record of the progressible {@link Object}s running on each
 	 * {@link Thread}.
 	 */
-	private static final ThreadLocal<ArrayDeque<ProgressibleObject>> progressibleStack =
+	private static final ThreadLocal<ArrayDeque<Task>> progressibleStack =
 		new InheritableThreadLocal<>()
 		{
 
 			@Override
-			protected ArrayDeque<ProgressibleObject> childValue(
-				ArrayDeque<ProgressibleObject> parentValue)
-		{
+			protected ArrayDeque<Task> childValue(ArrayDeque<Task> parentValue) {
 				// Child threads should be aware of the Tasks operating on the parent.
 				// For example, a progressible Object might wish to parallelize one of
 				// its stages; the child threads must know which Task to update
@@ -82,7 +80,7 @@ public final class Progress {
 			}
 
 			@Override
-			protected ArrayDeque<ProgressibleObject> initialValue() {
+			protected ArrayDeque<Task> initialValue() {
 				return new ArrayDeque<>();
 			}
 		};
@@ -138,15 +136,25 @@ public final class Progress {
 	 */
 	public static void complete() {
 		// update completed task
-		ProgressibleObject completed = progressibleStack.get().pop();
-		if (!completed.task().isComplete()) {
-			completed.task().complete();
+		Task t = progressibleStack.get().pop();
+		if (!t.isComplete()) {
+			t.complete();
 			// ping relevant listeners
-			pingListeners(completed);
+			pingListeners(t);
 			if (progressibleStack.get().peek() != null) {
 				pingListeners(progressibleStack.get().peek());
 			}
 		}
+	}
+
+	/**
+	 * Adds a new NOP {@link Task} to the current execution chain, used to ignore
+	 * {@link Progress} calls by a progressible object. This NOP task will handle
+	 * all {@code Progress} calls on this {@link Thread} between the time this
+	 * method is invoked and the time when {@link Progress#complete()} is called.
+	 */
+	public static void ignore() {
+		progressibleStack.get().push(IGNORED);
 	}
 
 	/**
@@ -180,13 +188,13 @@ public final class Progress {
 		var parent = deque.peek();
 		if (parent == null) {
 			// completely new execution hierarchy
-			t = new Task(description);
+			t = new Task(progressible, description);
 		}
 		else {
 			// part of an existing execution hierarchy
-			t = parent.task().createSubtask(description);
+			t = parent.createSubtask(progressible, description);
 		}
-		deque.push(new ProgressibleObject(progressible, t));
+		deque.push(t);
 		// Ping Listeners about the registration of progressible
 		pingListeners(deque.peek());
 	}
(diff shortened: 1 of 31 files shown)


scijava-ops-image/pom.xml
```diff
@@ -330,5 +330,11 @@
 			<artifactId>scifio</artifactId>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
+            <groupId>org.scijava</groupId>
+            <artifactId>scijava-progress</artifactId>
+            <version>0-SNAPSHOT</version>
+            <scope>compile</scope>
```
gselzer: This scope is unnecessary, and the version should probably be `${project.version}`

scijava-ops-image/src/main/java/org/scijava/ops/image/adapt/LiftComputersToRAI.java
```diff
@@ -71,8 +73,19 @@ private LiftComputersToRAI() {
...
-				.forEachPixel((i1, out) -> computer.compute(i1, out));
+				.forEachChunk((chunk) -> {
+					var counter = new LongType(0L);
+					chunk.forEachPixel((i1, out) -> {
+						computer.compute(i1, out);
+						counter.inc();
+					});
+					Progress.update(counter.getIntegerLong(), t);
+					return null;
+				});
```
gselzer: Let's either add this all over, or not at all. I'd prefer the latter, to get this PR merged faster, although maybe we should copy this code into the progress module somehow for a test that ensures correct behavior.

scijava-progress/src/main/java/org/scijava/progress/Progress.java
```diff
@@ -159,7 +157,7 @@ public static void complete() {
 	 *          progress.
 	 */
 	public static void register(final Object progressible) {
-		register(progressible, progressible.toString());
+		register(progressible, progressible.toString(), false);
```
gselzer: Ideally, `silent` wouldn't be passed along here - we could instead make `register` *not* ping listeners, and return the `Task`. Then, in `AbstractRichOp`, we could set the `Task` to silent, and then ping the listeners somehow?

scijava-progress/src/main/java/org/scijava/progress/Progress.java
```diff
@@ -174,19 +172,37 @@ public static void register(final Object progressible) {
...
-			t = new Task(description);
+			t = new Task(progressible, description, silent);
 		}
 		else {
 			// part of an existing execution hierarchy
-			t = parent.task().createSubtask(description);
+			t = parent.createSubtask(progressible, description, silent);
 		}
-		deque.push(new ProgressibleObject(progressible, t));
+//		deque.push(new Task(progressible, t));
```
gselzer: Delete commented out line

scijava-progress/src/main/java/org/scijava/progress/Progress.java
```diff
@@ -195,18 +211,23 @@ public static void register(final Object progressible,
...
 		// Ping object-specific listeners
-		List<ProgressListener> list = progressibleListeners.getOrDefault(o.object(),
-			Collections.emptyList());
+		List<ProgressListener> list = progressibleListeners.getOrDefault( //
+			task.progressible(), //
+			Collections.emptyList() //
+		);
 		synchronized (list) {
-			list.forEach(l -> l.acknowledgeUpdate(o.task()));
+			list.forEach(l -> l.acknowledgeUpdate(task));
```
gselzer, line 218: Could be a method reference

scijava-progress/src/main/java/org/scijava/progress/Progress.java
```diff
@@ -237,8 +257,18 @@ public static void update() {
...
+
+	/**
+	 * Updates the progress of the current {@link Task}, pinging any interested
+	 * {@link ProgressListener}s.
+	 *
+	 * @see Task#update(long)
+	 */
+	public static void update(final long numElements, final Task task) {
+		task.update(numElements);
+		pingListeners(task);
```
gselzer, line 266: Man, I wish that the `Task` itself had `pingListener`s. It seems smelly to me that `Progress.update` just calls `Task.update()`, but then does more that wouldn't happen if you called `Task.update()`. The whole point of this PR is to enable `Task` access, but then you won't receive updates by calling `Task.update()`...

scijava-progress/src/main/java/org/scijava/progress/Task.java
```diff
@@ -79,21 +82,38 @@ public class Task {
 
 	/** True iff {@link Task#max} has been defined for the current stage */
 	private boolean updateDefined = false;
+	private final boolean silent;
 
 	/** String identifying the task */
 	private final String description;
 
 	/** Computation status as defined by the task */
 	private String status = "Executing...";
```
gselzer: Consider `progressible.toString()` as the initial value?

gselzer (commented):
Nice change! Hopefully it makes updating progress easier

scijava-ops-image/pom.xml
gselzer: Fixed

scijava-progress/src/main/java/org/scijava/progress/Progress.java
gselzer: Deleted

scijava-progress/src/main/java/org/scijava/progress/Progress.java
gselzer: Actually, it can't be :sweat_smile:

scijava-progress/src/main/java/org/scijava/progress/Task.java
gselzer: That doesn't make sense since that is what the description is for, but I did make the default status the empty string.

scijava-progress/src/main/java/org/scijava/progress/Progress.java
gselzer: This is avoided through usage of `Progress.ignore()`

scijava-progress/src/main/java/org/scijava/progress/Progress.java
gselzer: Maybe I'll make another issue for this, because it's unrelated to this PR.

PR #132, opened 2024-03-05
Add Hint to control writing to Op history

gselzer:
This PR modifies the behavior of each `AbstractRichOp` to *only* record their executions within the `OpHistory` if a new Op hint `"History.RECORD"` is set. The previous mechanism for recording history, a field `boolean AbstractRichOp.record`, has been removed in favor of adopting the `Hints` paradigm.

The more important change is that recording  to the `OpHistory` is now **opt-in**, where it was opt-out before. This provides performance benefits (anecodotal, found in the process of unrelated benchmarking)

An additional change that is worth scrutiny is the fact that the aspect of History reporting is no longer a mutable property of a `RichOp`. Note there is (should be) nothing stopping users from re-matching an Op with this `Hint`.


hinerm:
My understanding was that one of the design goals of SJ Ops was to record what was done to facilitate reproducibility. I believe an opt-in mechanism practically will mean it will be unused.

Were the benchmarking problems you ran into arising from numerous history recording events stemming from a single Op call? If so, perhaps we just need to re-evaluate what/when history is actually being recorded... e.g. if there was only one recording event per call would that still incur performance hits?

PR #136, opened 2024-03-14
WIP add normalize op

hinerm:


hinerm:
I started this branch to work on https://github.com/scijava/scijava/issues/197

However, this implementation of normalization is not really what I wanted. In my sample data, the 16-bit histogram looks like this:
image

Really what I want is to normalize in this actual min-max range. Normalizing in the input (16-bit) range to the output (8-bit range results in all of these values being mapped essentially to 0-4ish which is.. not great.

I don't think using a single type-based op and lifting will be appropriate because we need to know the range. We would need a new Op that operates on images and has a dependency on min/max ops and normalize ops.

Normalize would also need to be updated to add signatures that accept input range parameters

hinerm:
Found the [`image.normalize`](https://github.com/scijava/incubator/tree/170852d2881d0a8eb5fa8245537583d4ceb062da/scijava-ops-image/src/main/java/org/scijava/ops/image/image/normalize) ops and they appear to work for my use case so I am going to just use those instead

PR #139, opened 2024-03-22
Use Anys, not wildcards, for converter output types

gselzer:
Wildcard types are problematic due to the restrictions on e.g. the number of bounds. `Any`s are more flexible in this regard, and better reflect our needs.

TODO:
* [x] Test changeset.

### DIFF
--- a/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/convert/Conversions.java
+++ b/scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/convert/Conversions.java
@@ -34,6 +34,7 @@ import org.scijava.function.Container;
 import org.scijava.function.Mutable;
 import org.scijava.ops.api.*;
 import org.scijava.ops.engine.BaseOpHints;
+import org.scijava.types.Any;
 import org.scijava.types.Nil;
 import org.scijava.types.Types;
 import org.scijava.types.inference.FunctionalInterfaces;
@@ -136,7 +137,7 @@ public final class Conversions {
 			return opt.get();
 		}
 		// Attempt 2: Computer with identity mutable output
-		opt = postprocessIdentity(info, request, preConverters, env);
+		opt = postprocessIdentity(info, request, preConverters, vars, env);
 		if (opt.isPresent()) {
 			return opt.get();
 		}
@@ -209,7 +210,8 @@ public final class Conversions {
 			postConverter, //
 			request.getOutType(), //
 			null, //
-			env //
+			env, //
+			vars //
 		));
 	}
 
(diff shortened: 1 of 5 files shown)


scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/convert/Conversions.java
```diff
@@ -443,13 +448,17 @@ private static void resolveTypes(Type source, Type dest,
...
+		if (t instanceof TypeVariable<?>) {
+			TypeVariable<?> tv = (TypeVariable<?>) t;
+			return Nil.of(new Any(tv.getBounds()));
+		}
 		var vars = new HashMap<TypeVariable<?>, Type>();
 		for (Type typeParam : typeParams) {
 			if (typeParam instanceof TypeVariable<?>) {
 				// Get the type variable
 				TypeVariable<?> from = (TypeVariable<?>) typeParam;
 				// Create a wildcard type with the type variable bounds
```
ctrueden, line 460: This comment is no longer correct.

scijava-ops-engine/src/main/java/org/scijava/ops/engine/matcher/convert/IdentityCollection.java
```diff
@@ -55,7 +55,7 @@ public class IdentityCollection<T, U extends T> implements OpCollection {
 	@OpHints(hints = { Conversion.FORBIDDEN,
 		BaseOpHints.DependencyMatching.FORBIDDEN })
 	@OpField(names = "engine.convert, engine.identity", priority = Priority.FIRST)
-	public final Function<U, T> identity = (t) -> t;
+	public final Function<T, T> identity = (t) -> t;
```
ctrueden, line 58: The parens are unnecessary; you can just write `t -> t`.

scijava-ops-engine/src/test/java/org/scijava/ops/engine/matcher/convert/ConversionTest.java
```diff
@@ -169,4 +171,43 @@ public void testConvertAnys() {
...
 
+	/**
+	 * An Op, written as a method, whose type variable has multiple bounds.
+	 * <p>
+	 * Note that, for the purposes of this test, the {@link Number} bound is
+	 * necessary even though it is not needed for the functionality of the test
+	 * Op.
+	 * </p>
+	 */
+	@OpMethod(names = "test.boundsConversion", type = BiFunction.class)
```
ctrueden, line 182: Once again I got confused by the name of this op. It's weird to me that we have test ops named after the *functionality being tested* rather than named for *what they do* like normal ops. In this case, the op `test.boundsConversion` is a static method named `foo` (which I find unhelpful in understanding) and it returns the max of two objects. So why not name it "max" instead? Especially since the functionality being tested here, multiple type bounds, is also a sort of comparison, which was the source of my momentary confusion: "What does `x > y ? x : y` have to do with *multiple* bounds?" I thought, "It's just *one* bound..."

ctrueden (commented):
This looks great! Very clean. I just have two tiny nitpicks, and one slightly less tiny nitpick. But nothing that should hold up a merge.

PR #140, opened 2024-03-26
Add custom Groovy lexer to sphinx/rtd

elevans:
This PR adds a `lexers.py` module that contains a custom Groovy lexer that adds "#@" lines as comments. To use this custom lexer simply define your code block like so:

```text
.. code-block:: scijava-groovy
```

This PR resolves the annoying errors with "#@" in groovy scripts and stops the error highlighting in our scijava ops docs.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -5,8 +5,11 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 
 # -- General configuration ---------------------------------------------------
-
+import sys
+import os
 from multiproject.utils import get_project
+
+sys.path.insert(0, os.path.abspath('.'))
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -56,6 +59,13 @@
 myst_heading_anchors = 4
 nb_execution_mode = "off"
 
+# -- Custom Lexers -----------------------------------------------------------
+
+from lexers import FijiGroovyLexer
```
ctrueden: Could we please call it `SciJavaGroovyLexer` instead? It really has nothing to do with Fiji specifically. The `#@` syntax is part of the SciJava scripting subsystem.

docs/conf.py
```diff
@@ -56,6 +59,13 @@
...
+
+def setup(app):
+    app.add_lexer('fijigroovy', FijiGroovyLexer)
```
ctrueden: And similarly, here I would name the lexer `scijava-groovy`. With dash for clarity, assuming dashes are allowed. Or else `scijava_groovy` or `sciJavaGroovy` if dashes cannot be used.

ctrueden (changes requested):
Thanks a lot for figuring this out! I would just request the name be "SciJava" instead of "Fiji" since that's the relevant layer here.

docs/conf.py
elevans: Done!

docs/conf.py
elevans: We can use dashes. The SciJava Groovy lexer alias is now `scijava-groovy`.

elevans (commented):
Thanks for the feedback @ctrueden! We can use this custom Groovy lexer with the `scijava-groovy` alias.

ctrueden (approved):
Thanks @elevans! LGTM

PR #142, opened 2024-04-01
Add OpenCV Use case

hinerm:


hinerm:
@elevans what do you think of this example?

elevans:
@hinerm I pulled the `opencv-use` branch and tried your example de-noise script, but it doesn't work. My output image looks the same as my input image (`sample_16bit_T24.png`) regardless of whatever setting I specify. Is there something more that I need to do to make this work?

edit: Just updating this for the record. This use case works when using Fiji with `jaunch`!

docs/ops/images/sample_16bit_T24.png
elevans, line 1: Think this file should live on `media.imagej.net` and when its ready `media.scijava.org` instead of the repo proper.

docs/ops/doc/examples/opencv_denoise.rst
```diff
@@ -0,0 +1,87 @@
+========================================
+Fast Non-Local Means Denoise with OpenCV
+========================================
+
+In this example we will use a denoise algorithm defined in an `external libray`_: OpenCV.
+
+This method progressively scans neighborhoods of an image looking for repeated search template. These
+repeated areas can then be averaged to eliminate gaussian noise, without the requirement of additional
+images for comparison.
```
elevans: Can you better explain what **strength**, **template size** and **search size** mean? What are the general consequences of the values? I don't think we need to be extremely detailed but it would be nice to have some context for the values.

elevans (changes requested):
Great example @hinerm! I it just needs some clarification on what the parameters do. It's too bad that for now this is limited to `uint8` outputs.

docs/ops/doc/examples/opencv_denoise.rst
hinerm: @elevans so the "external library" link goes to the OpenCV example which itself links to [the paper](https://www.ipol.im/pub/art/2011/bcm_nlm/article.pdf) defining the algorithm. I could recapitulate what's said in the OpenCV example, or point directly to the paper, or emphasize the OpenCV example for background reading.. any preference?

elevans: I think we should site the source paper (I used a super script with a number and then added a key to the bottom of the document) and recapitulate the parameter descriptions from the example:

```text
    h : parameter deciding filter strength. Higher h value removes noise better, but removes details of image also. (10 is ok)
    hForColorComponents : same as h, but for color images only. (normally same as h)
    templateWindowSize : should be odd. (recommended 7)
    searchWindowSize : should be odd. (recommended 21)
```

But I think we can do better here and also be more explicit about what these parameters are (I renamed "template"  to "patch size":

```text
- Filter Strength (h): Controls the decay in patch weights as a function of "distance" between patches. The "distance" between patches is a measure of how similar they are. Large filter strength values allows more distant (i.e. dissimilar) patches to have more influence on the denoise output.
- patch size: The size of the patches/blocks from the input image to be compared. Patch sizes should be odd (default=7).
- search size: The size of the area in the input image to search for similar patches to compare. Search sizes should be odd (default=21)
```

Thoughts?

elevans: Here's [scikit-image](https://scikit-image.org/docs/stable/auto_examples/filters/plot_nonlocal_means.html)'s implementation of the same algorithm.

hinerm: @elevans 

>Thoughts?

This sounds good.. can you push those changes?

elevans: @hinerm I added some improvements to your denoise use case. Let me know what you think. If you like the changes, then I think this branch is ready to merged.

elevans (approved):

PR #143, opened 2024-04-02
Miscellaneous cleanups

ctrueden:
This patchset cleans up various things:

* The list of contributors in the POM.
* The mailmap, for better git behavior.
* The naming of LiftNeighborhoodComputersToRAI.
* The description of each module in the suite.
* The use of confusing and/or obsolete terminology.

And it updates the license headers as well.


scijava-common3/pom.xml
```diff
@@ -11,8 +11,8 @@
 
 	<artifactId>scijava-common3</artifactId>
 
-	<name>SciJava Common 3 </name>
```
gselzer, line 14: Hmm, we should talk about this.

I thought that we were intending to version library incarnations, and it seems like it would be confusing to not name them differently. What do you think about this change?

scijava-ops-api/pom.xml
```diff
@@ -12,7 +12,7 @@
 	<artifactId>scijava-ops-api</artifactId>
 
 	<name>SciJava Ops API</name>
-	<description>SciJava Operations API: Outward-facing Interfaces used by the SciJava Operations framework.</description>
```
gselzer, line 15: I like invoking "user" connotations here, what do you think about "The user interface for the SciJava Ops framework"?

gselzer: Or something similar

scijava-struct/pom.xml
```diff
@@ -12,7 +12,7 @@
 	<artifactId>scijava-struct</artifactId>
 
 	<name>SciJava Struct</name>
-	<description>A lightweight framework for collecting Members</description>
```
gselzer, line 15: "Structured data` feels awfully broad to me. The `README` uses the term "typed fields" - what do you think about that?

Regardless should make sure the `pom` and the `README` are consistent

gselzer (commented):
Just a couple small questions about POM descriptions!

scijava-common3/pom.xml
ctrueden: My rationale is that I modeled the naming after projects like Apache Commons Math. They still call it "Apache Commons Math" as a name, even though the artifact is commons-math4 and the package also is `org.apache.commons.math4`. It's just the name that doesn't have the `4` in it.

Now in the case of SciJava Common, the scope of the library is drastically reduced from what it was before. So anyone thinking that this new scijava-common3 could possibly replace the previous scijava-common is in for a rude shock if they try it. But I'm not super concerned about that.

Anyway, I don't feel super strongly, if you'd rather make the name "SciJava Common 3" then I guess we can do that. Or should we do "SciJava Common3"? I wasn't sure, so I opted to just ditch the "3". :sweat_smile:

scijava-ops-api/pom.xml
ctrueden: The term "user interface" implies GUI, so I'd rather not. Something with "application programming interface" or just "API" would be OK with me though. GitHub won't show me what I actually changed the description to here, which is frustrating.

scijava-struct/pom.xml
ctrueden: The term "typed fields" would be OK with me. From Wikipedia: "A record is a collection of fields, possibly of different data types, typically in a fixed number and sequence."

scijava-ops-api/pom.xml
gselzer: "User API"?

scijava-common3/pom.xml
gselzer: Gotcha, didn't realize that was the precedent. Let's just go with it then.

ctrueden:
@gselzer
For scijava-struct, how about "A library for building and introspecting structs."?
For scijava-ops-api, how about "Public API for using SciJava Ops."?
For scijava-common3, I defer to your judgment how you would prefer it be named. My weak preference is just "SciJava Common" for the name, without the 3.

ctrueden:
On second thought, for scijava-ops-api, I think I prefer "The public API of SciJava Ops."

ctrueden:
I pushed revised descriptions for scijava-ops-api and scijava-struct. I did not change scijava-common3 yet. @gselzer Let me know how you want to proceed.

gselzer:
> I pushed revised descriptions for scijava-ops-api and scijava-struct. I did not change scijava-common3 yet. @gselzer Let me know how you want to proceed.

Let's just go with SciJava Common, I didn't realize that was precedent from e.g. Apache Commons Math

gselzer:
Thanks for making all these changes!

ctrueden:
@gselzer OK to merge? Or are further changes needed?

gselzer:
I think it looks great! Let's merge!

PR #145, opened 2024-04-04
Fix deconvolution namespace and quality test

elevans:
This PR fixes and refactors the deconvolution namespace :tada:! I compared deconvolution outputs from SciJava Ops with ImageJ Ops to ensure that outputs from both frameworks are the same. This PR also includes an example document with the Groovy and Python script example for Fiji.

A lingering question is whether we should include some pretty output pictures of the deconvolution result like we do in the [PyImageJ docs](https://py.imagej.net/en/latest/Deconvolution.html). I'm a little hesitant to start adding screen shots and other pictures to the repo as it might inflate it. Thoughts?


docs/ops/doc/examples/deconvolution.rst
```diff
@@ -0,0 +1,139 @@
...
+                                                                numericalAperture,
+                                                                wavelength,
+                                                                riSample,
+                                                                riImmersion,
+                                                                lateral_res,
+                                                                axial_res,
+                                                                pZ,
+                                                                FloatType()).apply()
+
+        # deconvole image
```
gselzer: deconvole

scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/AccelerationState.java
```diff
@@ -0,0 +1,49 @@
...
+ * #L%
+ */
+
+package org.scijava.ops.image.deconvolve;
+
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.img.Img;
+import net.imglib2.type.numeric.RealType;
+
+public class AccelerationState<T extends RealType<T>> {
```
gselzer, line 45: Some javadoc would be greatly appreciated

scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/VectorAccelerator.java
```diff
@@ -56,115 +53,102 @@
  * @param <T>
  * @implNote op names='deconvolve.accelerate', priority='0.'
  */
-public class VectorAccelerator<T extends RealType<T> & NativeType<T>> implements
-	Inplaces.Arity1<RandomAccessibleInterval<T>>
+public class VectorAccelerator<T extends RealType<T>> implements
```
gselzer, line 56: This Op is interesting now. There's a data structure, *not exposed by the module*, which appears as the input to an Op that users *could* (even though they shouldn't) call. This effectively makes the Op uncallable when JPMS is enforced (note that for now, in the scripts, it is fine). 

@ctrueden @elevans any opinions on this situation? Here are some initial options I see:
* Make this functionality into something that isn't an Op, possibly make it part of the `AcceleratorState`.
* Add a Hint where Ops can declare themselves as hidden.
* Leave it

gselzer (changes requested):
These were all great changes, thank you so much for spearheading them @elevans !

scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/VectorAccelerator.java
ctrueden: > There's a data structure, not exposed by the module, which appears as the input to an Op that users could (even though they shouldn't) call.

Arghhhhhhh :sweat_smile: Very good point.

Given the short time frame till the paper, my vote is "Leave it" and file an issue. Any user confusion about it is likely to be near zero—although if someone complains, we can expedite addressing it somehow.

elevans: Thanks @gselzer for explaining the finer details of this issue. My short term vote is to "leave it" as well for the same reason that Curtis has, we just don't have time. Long term, I'm leaning towards demoting the `VectorAccelerator` from an Op to a helper class.

gselzer: In that case let's file an issue. My only concern is that removing an Op is *technically* breaking backwards compatibility, but it's probably fine :wink:

elevans: Hmmm well I'm only leaning that way, so I'm not advocating we remove it right now :nerd_face:. For now we will leave this.

scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/AccelerationState.java
elevans: Done 49cf29dfa0e94c6e397f6cd35ad2f50d6b892981.

docs/ops/doc/examples/deconvolution.rst
elevans: Oops! fixed :nerd_face: b73b91bba1cc3456567e6ef77fc70cf16426b83f

elevans:
@gselzer Thanks for the review! I fixed the typos and added the javadoc to `AccelerationState.java`. Lastly, we decided to leave the `VectorAccelerator` as is for now.

elevans:
Now that we have [media.imagej.net/scijava-ops](https://media.imagej.net/scijava-ops/) I want to add an expected results image to the deconvolution example. So for now, please hold off on merging this PR. I will get this done by tomorrow.

elevans:
I updated the decon example. Its ready for a last review. @gselzer, what do you think?

hinerm:
@ctrueden @elevans was there a particular reason the `AccelerationState` has pythonic-style field names and/or public parameter access?

Any reason not to a) refactor to getters/setters or b) make the `AccelerationState` protected?

ctrueden:
@hinerm The fields are final so it is fine for them to be public.

The `AccelerationState` class needs to be public because it is a parameter type for an Op. Otherwise there would be an Op in the environment that users cannot actually call due to lack of type visibility. Actually, that is already the case due to the package not being exported... but making the class non-public would compound the issue.

ctrueden:
Err crap, they *aren't* final! Hmm...

ctrueden:
How would you rather proceed?

A) Fix the AccelerationState API and export the package?

OR

B) Make the acceleration thingie not an Op?

I think having private Ops is a bit strange.

OR

C) Decide on and document exactly when Ops should be allowed to be private and why...

hinerm:
>A) Fix the AccelerationState API and export the package?

@ctrueden I'm inclined to just do this. I added getters/setters locally and rebuilt. Tests and the script still seem to work.

hinerm (approved):
The decon example worked locally on my system, so LGTM.

ctrueden:
I am seeing the same test failure on the main branch:
```
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.015 s <<< FAILURE! -- in org.scijava.ops.image.transform.project.project.ProjectTest
  Error:  org.scijava.ops.image.transform.project.project.ProjectTest.testProjector -- Time elapsed: 0.012 s <<< ERROR!
  java.lang.ArrayIndexOutOfBoundsException
```
Is this the race condition with `LoopBuilder`?

scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/AccelerationState.java
```diff
@@ -44,15 +44,67 @@
...
+	private Img<T> gk;
+	private Img<T> gkm1;
+	private double accelerationFactor = 0.0f;
+
+	public AccelerationState(RandomAccessibleInterval<T> ykIterated) {
+		this.ykIterated = ykIterated;
+	}
+
+	public RandomAccessibleInterval<T> ykIterated() {
+		return ykIterated;
```
ctrueden, line 60: This field is final, so does not need a getter...

PR #147, opened 2024-04-11
Improve the SciJava Ops benchmarks

### LINKED ISSUE #210: OpBuilder: inplace methods don't preserve hints
See 
https://github.com/scijava/incubator/blob/6157f6e052f5bf9ad3936dc1b5fda5ca6e119f01/scijava-ops-api/src/main/java/org/scijava/ops/api/OpBuilder.java#L2435

ctrueden:
* Make the underlying op implementation as trivial as possible, while still preventing the JIT from optimizing away method calls completely.
* Merge all benchmarking classes into a single `Benchmarks.java`, to avoid repeatedly benchmarking the same thing across the classes.
* Add static/non-Ops benchmarks that also perform adaptation, conversion, and both, so that we can compare apples to apples there.
* Update the Python script accordingly to consume the revised benchmarking output, and generate the same charts as before, with a couple of additional bars where appropriate.
* Switch the unit of measure from milliseconds to microseconds.
* Fix a couple of trivial issues with the Benchmarks doc page.


ctrueden:
The question of why cached matches are slower than non-cached ones is still open, and I plan to investigate soon. But that should not hold up the merger of this PR, which I feel is a step forward for the benchmarks regardless.

docs/ops/doc/index.rst
```diff
@@ -22,6 +22,7 @@ The combination of these libraries allows declarative image analysis workflows,
 
    CallingOps
    SearchingForOps
+   ScriptingInFiji
```
gselzer: This file is now duplicated in the list

docs/ops/bin/benchmark.sh
```diff
@@ -65,4 +65,4 @@ fi
 
 echo
 echo '=== GRAPHING THE RESULTS ==='
-"$pythonPath" graph_results.py
+#"$pythonPath" graph_results.py
```
gselzer: Should this be uncommented?

scijava-ops-benchmarks/src/main/java/org/scijava/ops/benchmarks/Benchmarks.java
```diff
@@ -0,0 +1,189 @@
...
+		state.env.unary("benchmark.increment").input(state.theByte).mutate();
+	}
+
+	/**
+	 * SciJava Ops, caching disabled.
+	 *
+	 * @param state the state between benchmarks
+	 */
+	@Benchmark
+	public void sjOpsNoCache(final BenchmarkState state) {
```
gselzer: It's a little strange that this benchmark and the one below all ignore the cache, but only this one has `NoCache` in the name. But it's probably fine.

gselzer (commented):
Super awesome! Couple small suggestions

gselzer:
@ctrueden it would also be nice to have a little section at the end of the RTD benchmark page with the combined graph - otherwise, the JSON you've included will be inaccessible.

hinerm (commented):
Looks good to me with Gabe's comments addressed and the caching investigation.

scijava-ops-benchmarks/src/main/java/org/scijava/ops/benchmarks/Benchmarks.java
ctrueden: I know, I don't like it. I was going to rename it to be more consistent, but forgot. I'll take care of it!

ctrueden:
Thanks! I've addressed all the feedback points:

* No more duplicate `ScriptingInFiji`
* No more commented-out python script execution step
* More consistent naming of the sjOps test methods

About adding the fifth "combined metrics" chart: it was already there, so I think all is well? Because I had slipped it onto the main branch a few days ago, before this more major reworking. :sweat_smile: 

Still working on updating the `Benchmarks.rst` to be correct, and additional adaptation tests, and caching investigation.

hinerm:
@gselzer I removed inplaces and made the test Op a `Computer` that's adapted to a `Function`.

However, the nullary computer does not match. The `testSciJavaOps` and `testConversion` tests fail:

```
org.scijava.ops.api.OpMatchingException: No match found! Perhaps you meant: 
benchmark.increment:
	- org.scijava.ops.benchmarks.OpsToBenchmark.incrementByte(byte[])
		 Increments a byte value.
		
	
		> mutable1 : @MUTABLE byte[]
			array containing the byte to increment
...

Request:
-	Name: "benchmark.increment", Types: org.scijava.function.Computers.Arity0<[B>
Input Types: 
		* byte[]
Output Type: 
		* byte[]
```

PR #148, opened 2024-04-16
Remove arity methods

hinerm:
Removes the `OpBuilder.arityX` methods and `OpEnvironment.*ary` methods.


scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -3,6 +3,32 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+*/
+
```
gselzer: There was some reason that @ctrueden had removed licenses in the `.vm` files, in this commit: https://github.com/scijava/incubator/commit/df940e262c4b08b29b27644758ffe5d746b9ac61

scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
```diff
@@ -881,20 +835,20 @@ public class OpBuilder {
...
 		public String help() {
-		    return env.help(new PartialOpRequest(opName, new Nil<?>[] {$inputTypesFromArgs.call($arity)}, type(out)));
+			return env.help(new PartialOpRequest(opName, new Nil<?>[] {$inputTypesFromArgs.call($arity)}, type(out)));
 		}
 
-        /**
-         * Finds all Ops matching the current partial OpBuilder request
-         * @return a {@link String} with a verbose entry for each Op satisfying the partial request
-         */
+		/**
```
gselzer, line 820: Is the indentation correct in the output?

scijava-ops-engine/src/test/java/org/scijava/ops/engine/OpBuilderTest.java
```diff
@@ -187,7 +187,7 @@ public void testArity1_IT_OU_matchI1() {
 		final double[] expected = { 1.0, 2.0, 3.0 };
 		final Inplaces.Arity1<double[]> op = //
 			name("test.mulArrays1_1") //
-				.arity1() //
+				//
```
gselzer, line 190: Can we delete these?

scijava-ops-engine/src/test/java/org/scijava/ops/engine/adapt/AdaptationTypeVariableCaptureTest.java
```diff
@@ -112,7 +112,7 @@ public Function<I, O> apply(Computers.Arity1<I, O> computer) {
 	public void testCapture() {
 		List<Byte> in = Arrays.asList((byte) 0, (byte) 1);
 		Object out = ops.op("test.adaptedCapture") //
-			.arity1() //
+			//
```
gselzer, line 115: Here's another empty comment - can we find and remove all of these?

scijava-ops-engine/src/test/java/org/scijava/ops/engine/adapt/lift/ComputerToIterablesTest.java
```diff
@@ -66,7 +66,7 @@ public void testComputer1ToIterables() {
 		final List<double[]> expected = //
 			Arrays.asList(new double[] { 1., 2., 3. });
 		ops.op("test.addArrays") //
-			.arity1() //
+			//
```
gselzer, line 69: Here's another

gselzer (approved):
Looks great!

scijava-ops-engine/src/test/java/org/scijava/ops/engine/OpBuilderTest.java
gselzer: I deleted!

scijava-ops-engine/src/test/java/org/scijava/ops/engine/adapt/AdaptationTypeVariableCaptureTest.java
gselzer: Removed!

scijava-ops-engine/src/test/java/org/scijava/ops/engine/adapt/lift/ComputerToIterablesTest.java
gselzer: Not anymore!

scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm
hinerm: @gselzer the `.vm` previously [mixed spaces and tabs](https://github.com/scijava/incubator/blob/6157f6e052f5bf9ad3936dc1b5fda5ca6e119f01/scijava-ops-api/templates/main/java/org/scijava/ops/api/OpBuilder.vm#L69-L73) but it seems the output was consistent both before and after this change

docs/ops/doc/examples/deconvolution.rst
```diff
@@ -80,7 +80,7 @@ SciJava Ops via Fiji's scripting engine with `script parameters`_:
         pZ = pZ * 1E-6
 
         // create the synthetic PSF
-        psf = ops.op("create.kernelDiffraction").arity9().input(psf_size,
+        psf = ops.op("create.kernelDiffraction").input(psf_size,
                                                                 numericalAperture,
```
ctrueden, line 84: These parameters are no longer indented in alignment with the first parameter `psf_size`.

ctrueden (commented):
Thanks so much for doing this, @hinerm. It looks awesome. I made one comment with a tiny nitpick I noticed. My only other thought is that the commit message doing the actual removal could explain in more detail why we are removing this. Here are the notes I hurriedly scrawled the other day:

<blockquote>

Why I believe the disadvantages of the `.arityX` syntax outweigh the advantages:
- compromises the core builder design in service to an IDE implementation detail
- it pollutes the syntax of every downstream scripting language, beyond only Java
- the additional complexity is not necessary complexity -- from a type safety perspective, it's redundant
- editing op builder expressions in an IDE becomes confusing when altering the number of input parameters naively
- experimenting with different op calls across arities is more burdensome because you have to keep tweaking the arity number
- new users must be taught the jargon "arity"
  - to be fair, it's also in the functional ifaces
    - but that only crops up when writing new ops, or when calling the op builder's op instance accessors `.function()`, `.computer()`, and `.inplace()`, which new users will typically not do
- addition of unary/binary/etc shortcuts reintroduces autocompletion cruft and terms are "out of order"
- gains one very specific transient benefit (helps new users only; and only when learning via IDE autocomplete) in exchange for a permanent general drawback (all users must deal with the hassle of the additional `.arityX` in every invocation in every language)

Why keep the `arityX` syntax:
- so that new people are less intimidated when using an IDE to build up queries
  - Counterpoint: They will instead be intimidated when they try to edit existing expressions and the number of parameters changes, which is a common occurrence given multiple Ops signatures (multiple implementations, and/or optional parameters).
  - Counterpoint: Why are we in Java? For type safety. We still achieve that without the `.arityX()` step.
- anyone who codes in an IDE and sees 48 options pop up is a "code smell", appears unorganized

If we don't keep `arityX`, how can we mitigate the autocompletion explosion of `in*` methods?
- Reduce the max support arity in the `OpBuilder` to 6 -- there are very few Ops >6 arity; we could change them to use config object (which we would then need to export -- thanks, JPMS)
  - scijava-ops-image/src/main/java/org/scijava/ops/image/filter/FFTMethodsLinearFFTFilterC.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/filter/convolve/ConvolveFFTC.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/filter/correlate/CorrelateFFTC.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/filter/correlate/CorrelateFFTF.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/create/Creators.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/PadAndRichardsonLucy.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/PadAndRichardsonLucyTV.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/PadAndRichardsonLucy.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/PadAndRichardsonLucyTV.java
  - scijava-ops-image/src/main/java/org/scijava/ops/image/deconvolve/RichardsonLucyC.java

Plan: gather opinions at the hackathon:
- which tradeoff is best?
  - code looking nicer / shorter?
  - or users not being terrified of the API?
- OR: other ideas to strike a balance here?
  - e.g. include "bare" input sigs up to ~4? input(a), input(a, b), input(a, b, c), input(a, b, c, d)

</blockquote>

PR #153, opened 2024-04-22
Add SciJava Ops with Python use case

elevans:
This PR adds a use case for using SciJava Ops with Python and without ImageJ!


hinerm (changes requested):
This is awesome @elevans.

What do you think about adding an `environment.yml` and referencing it as a starting point from `scyjava.rst`?

We already have one in `scijava-ops-tutorial`... could re-use that one if it's sufficient or add a new one?

elevans:
Great idea! I've added .yml file config you can use to the use case text.

hinerm:
Awesome. I think this just needs someone to reproduce it and then can be merged and added to the paper. @gselzer did you want to reproduce? Otherwise I can.

hinerm:
Thanks again @elevans!

I wanted to note that, oddly, when I pasted the full script into an interpreter I got this output:
image

But if I save the script and run it with `python script.py` I get appropriate output:
image

To me this shows that the script works and if we want to debug the interpreter thing we can do so separately, unless anyone has an immediate idea of why it might happen.

@gselzer are you comfortable with that for inclusion in the paper? If so we can merge

hinerm:
@elevans I went ahead and debugged. I think there is something on my system specifically with pasting code? 
image

so maybe let's just not worry about it 🙃

hinerm (approved):

PR #154, opened 2024-04-22
Append color and scale bars to FLIM example

gselzer:
This PR provides additional color and scale bars to the FLIM example output described on ops.scijava.org. Unfortunately, the mechanism for doing this is pure ImageJ, which is why I decided to partition the entire use case into two scripts, one with pure SciJava Ops and the other with pure ImageJ.

TODO: Before merging, delete unused FLIM figures from media.imagej.net

### DIFF
--- a/docs/ops/doc/examples/example_flim_analysis.rst
+++ b/docs/ops/doc/examples/example_flim_analysis.rst
@@ -6,10 +6,10 @@ In this example we will use SciJava Ops within Fiji to perform `FLIM`_ analysis,
 
 .. image:: https://media.imagej.net/scijava-ops/1.0.0/flim_example_input.gif
     :width: 49%
-.. image:: https://media.imagej.net/scijava-ops/1.0.0/flim_example_pseudocolored.png
+.. image:: https://media.imagej.net/scijava-ops/1.0.0/flim_example_pseudocolored_annotated.png
     :width: 49%
 
-We use a sample of `FluoCells™ Prepared Slide #1`_, imaged by `Jenu Chacko`_ using `Openscan-LSM`_ and SPC180 electronics with multiphoton excitation and a 40x WI lens.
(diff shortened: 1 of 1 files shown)


docs/ops/doc/examples/example_flim_analysis.rst
```diff
@@ -36,17 +36,15 @@ Script execution requires a number of parameters, which may be useful for adapti
 | Bin Kernel Radius                    | 1     |
```
hinerm, line 36: Any reason not to bake all of these as default values in the script? You're already doing it for lifetime axis.. I don't think it takes away from reusability on other data and will make it overall more foolproof for people who want to reproduce this example.

docs/ops/doc/examples/example_flim_analysis.rst
```diff
@@ -151,17 +142,91 @@ In the panels below, we show script execution with computation restricted to the
...
+---------------------------------
+
+To attach quantitative meaning to the pseudocolored image, we must add calibration and scale bars. We add each within ImageJ using a second SciJava script using the following parameters. Note that ``tLow`` and ``tHigh`` are outputs from the previous script.
+
+With 130 microns across each direction in the dataset, we have 512/130~3.938 pixels per micron.
+
++--------------------------------------+---------------------+
+| Parameter                            | Value               |
++======================================+=====================+
+| tLow                                 |  1.87000572681427   |
```
hinerm, line 162: I would also add these as default values for this script. When I first pasted the values in they were rounded to 2, 2 and 4 in the UI🙄 but subsequent runs showed full precision.

hinerm (changes requested):
I was able to reproduce the output and add color bars. Thanks @gselzer!

I personally think it's worth adding default values to the scripts but other than that I don't see any changes necessary.

One note that we have no control over: after adding scale bars, if I click on a different active window (e.g. the numerical output window) then the bar text... disappears:
image

I only mention this in case others run into it. Re-activating the pseudocolored output and re-running the color bar script returned the text. Hooray.

PR #158, opened 2024-05-01
Update links to use media.scijava.org

ctrueden:


gselzer (commented):
@ctrueden thanks for doing this!

The only barrier I see to merging this is rebasing over the latest `main` and doing it again (because there will be new links to `media.imagej.net` on e.g. the papers page)

ctrueden:
Ahh sorry @gselzer, I forgot to pull before doing this. Redone now.

